PR #2, opened 2019-06-26

jaspindersingh83:


Exercise_2.java
```diff
@@ -9,6 +9,7 @@
         StackNode(int data) 
         { 
             //Constructor here 
+            int i = 10;
```
super30admin: You just added a i = 10. How we are going to use it?

PR #3, opened 2019-06-27
exercise1

sarang4488:
Implemented Exercise 1


super30admin:
I really appreciate the way you used Exceptions in Exercise_1. Implementation of other three exercises was equally good. Well Done !! Sarang.

PR #7, opened 2019-06-29

Ameya145:


LinkedList.java
```diff
@@ -4,7 +4,7 @@
 // a Singly Linked List 
 public class LinkedList { 
   
-    Node head; // head of list 
+    static Node head; // head of list
```
super30admin: Could you please state the reason for head being static here?

super30admin:
Well Done!! You have completed the assigned tasks correctly.
Your solutions suggest that you seem to be good at Stacks and Linked List.
Suggestion:
-Try to Comment your approach to the solution along with the code.

LinkedList.java
Ameya145: Oh! There is no reason as such. I was getting warnings from Eclipse regarding this, hence I changed. 
Making it static did not affect my solution as such.

super30admin: Would it work if you try to make more than one Linked Lists using this class ?

PR #8, opened 2019-06-29
Completed Precourse1

gauravj09:


Exercise_2.cpp
```diff
@@ -19,21 +20,43 @@ StackNode* newNode(int data)
...
 void push(StackNode** root, int data) 
 { 
     //Your code here 
+    StackNode* node = newNode(data);
+    node->next = *root;
+    *root = node;
 } 
-  
+
+//Why does pop method need a parameter since Stack is LIFO?
```
super30admin: Can you actually do this without having access to root ?

gauravj09: That makes sense. I was a little confused initially. Thank you for the response

super30admin:
Bravo! You have coded the solutions very nicely.
Keep Going. Good Luck for next Exercise.
Moreover, the other positive we found is that the way you raised doubt shows your interest in Data Structures.

PR #9, opened 2019-06-29
I added int

PrachiJani13:


PrachiJani13:
Implemented Exercise_1, Exercise_2, Exercise_3, Exercise_4

PR #10, opened 2019-06-29
PreCourse_1 Review

Hariae:
Renamed files to match JAVA classes.


GFG.java
```diff
@@ -1,59 +1,87 @@
...
+       
+    /*function to insert element in binary tree */
+    static void insert(Node temp, int key) 
+    { 
+
+        // Do level order traversal until we find 
+        // an empty place and add the node.  
+        if(temp == null){
+            temp = new Node(key);
+            root = temp;
```
super30admin: Can you explain reason to line 40 and 41? 
Is this actually required?

Hariae: I added with intention to create root node if the tree is empty. Now realized since the insert method doesn't return anything we can't create a root node that way.

super30admin: Yeah!! Exactly.
Good to see that you got that yourself.

PR #11, opened 2019-06-29
Completed problem 1

lavanganji:


super30admin:
Very nice work lavan!!
The efforts you put in commenting your approach along with the code are worth praising.

PR #23, opened 2019-07-29
Please review 

Rakshitha2020:
@kaurramanpreet

### DIFF
--- a/Exercise_3.py
+++ b/Exercise_3.py
@@ -1,1 +1,63 @@
 
+class Node:
+
+    def __init__(self,data):
+        self.data=data
+        self.next=None
+
+class LinkedList:
+    def __init__(self):
+        self.head=None
+
+
+    def push(self,data):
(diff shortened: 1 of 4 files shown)


Exercise_3.py
```diff
@@ -1 +1,63 @@
...
+
+class LinkedList:
+    def __init__(self):
+        self.head=None
+
+
+    def push(self,data):
+        newnode=Node(data)
+        newnode.next=self.head
+        self.head=newnode
```
super30admin: can you give an explanation of your push method?

Rakshitha94: every new node is added as self.head and all the new elements are added in front of the old add. In other words, the new element is pointed to the old head and then the new node becomes the new self.head

kaurramanpreet: Good
from now on try to give the explanation to your code using comments
this will help you alot

PR #25, opened 2019-07-30
[Sweta Kumari] Exercise_1.java file of PreCourse1

SwetaK951:


kaurramanpreet:
Complete other questions as well

SwetaK951:
Updated the second question as well.

PR #31, opened 2019-08-19
Completed Pre Course 1 Problems

pradhan1234:
Note: Had to remove public, as the problem file names and class names were not consistent.

Problem 1 was already reviewed by @kaurramanpreet here #27 

Please review other 3 problems


kaurramanpreet:
Great work!!

PR #36, opened 2019-10-05
PreCourse-1 in python

Arihant1467:


kaurramanpreet:
Very well Done Arihant
every exercise is absolutely correct!!

now onwards never forget to mention the 4 points:

1 space time complexity
2 problem if you faced any while attempting the problem
3 small explanation for your approach
4 did your code run on leetcode or not!!

PR #45, opened 2019-10-07
Pre course 1 homework

vramanathan:
Attempt in solving home work


vramanathan:
completed pre course 1 homework (i think(

PR #54, opened 2019-11-12
 Completed 4 exercises

Nsharma335:
Precourse-1 completed. Please review.

### DIFF
--- a/Exercise_1.java
+++ b/Exercise_1.java
@@ -1,44 +1,64 @@
-class Stack { 
-    static final int MAX = 1000; 
-    int top; 
-    int a[] = new int[MAX]; // Maximum size of Stack 
-  
-    boolean isEmpty() 
-    { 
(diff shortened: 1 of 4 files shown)


Exercise_1.java
```diff
@@ -1,44 +1,64 @@
...
+		a = new int[MAX];
+	}
+
+	boolean push(int x) throws Exception {
+		/* Check for stack Overflow */
+		if (top == MAX)
+			throw new Exception("Stack Overflow");
+
+		top++;
+		a[top] = x;
```
sehaj-grover: please check the if condition for overflow, top == Max-1
if we compare with max at 999 it will get overflow for the next push

Nsharma335: You are right, I updated the file.
Thanks.

sehaj-grover:
Great!

PR #73, opened 2019-12-19
Completed PreCourse_1 - Please review

preranas20:


Exercise_2.java
```diff
@@ -9,30 +20,62 @@
...
     } 
     
 	
     public boolean isEmpty() 
     { 
         //Write your code here for the condition if stack is empty. 
+        if(root == null)
+            return false;
+        return true; 
     }
```
sehaj-grover: Please review this method 
If root is null, that implies that the list is empty and should return true

preranas20:
Made the changes to Excercise_2.java

sehaj-grover:
Very good

PR #76, opened 2019-12-19
PreCourse_1: Submission by Bhushan

BhushanVasisht:


Exercise_1.java
```diff
@@ -1,33 +1,84 @@
...
     boolean push(int x) 
     { 
         //Check for stack Overflow
         //Write your code here
+        //check if top is equal to MAX
+        if(top == MAX)
+        {
+            System.out.println("Stack Overflow");
+            return false;
+        }
```
sehaj-grover: This condition will given an exception at index already at 999 you will try to add an element to top = 1000 
The if statement should be for top == MAX-1

BhushanVasisht: Yes. Thanks for the review. Have committed the changes to the repo

sehaj-grover:
Well done

PR #78, opened 2019-12-20
Precourse1 

nehabindle:


Stack.java
```diff
@@ -0,0 +1,75 @@
...
+    } 
+
+    Stack() 
+    { 
+        top = -1;
+    } 
+  
+    boolean push(int x) 
+    { 
+    	if(top >- (MAX -1))
```
sehaj-grover: Please recheck this condition
Most probably a typo though

PR #79, opened 2019-12-23
Completd 1,2,3

ghost:


Exercise_1.java
```diff
@@ -1,44 +1,65 @@
...
+	    } 
+
+	    Exercise_1() 
+	    { 
+	        top=-1;
+	    } 
+	  
+	    boolean push(int x) 
+	    { 
+	        if (top>=MAX){
```
sehaj-grover: This condition will give an exception because the last index for the array is MAX-1, So if the top is equal to MAX-1, we will get an exception

sehaj-grover (commented):
Please review my comment on exercise 1 and for exercise 2 (implementation stack using a linked list), We can add and remove the element on the head itself i.e. in push add the element as root and at pop remove it from the root instead of traversing till the end of the list

PR #80, opened 2019-12-25
Completed PreCourse 1- Please review

ashu20red:


Exercise_1.java
```diff
@@ -1,33 +1,73 @@
...
-        //Initialize your constructor 
+        top = -1;
     } 
   
     boolean push(int x) 
     { 
-        //Check for stack Overflow
-        //Write your code here
+        //Checking for stack Overflow
+        if(top >= MAX - 1)){
```
sehaj-grover: extra closing bracket

Exercise_1.java
```diff
@@ -1,33 +1,73 @@
...
+            System.out.println("Stack overflow");
+            return false;
+        }
+        else{
+            a[top++] = x;
+            System.out.printl(x);
```
sehaj-grover: println typo

Exercise_1.java
```diff
@@ -1,33 +1,73 @@
...
     { 
-        //Check for stack Overflow
-        //Write your code here
+        //Checking for stack Overflow
+        if(top >= MAX - 1)){
+            System.out.println("Stack overflow");
+            return false;
+        }
+        else{
+            a[top++] = x;
```
sehaj-grover: This will give you an exception at the very first insertion because top++ means that you are trying to add element at -1 index and then increment top 
Instead it should be ++top, which means that we are initially incrementing -1 to 0 and then adding the element at 0th index

ashu20red:
Hi,
Thank you so much for your review.

I will correct it and submit asap


Thanks.
Regards,
Ashwin Saxena

sehaj-grover:
Very good

PR #87, opened 2020-01-27
Finished PreCourse_1, Please review

sowmyadvn:


sowmyadvn:
> Great job with this sprint Sowmya
> Quick tips
> 
> 1. Include time and space at top of your file
> 2. Write what were the challenges that you face which we can address and help you with

Thanks for reviewing @super30admin. I just added time complexity of algorithms. One question I have is regarding time complexity of insert in binary tree. I assumed it is O(N) here because, in worst case, only the last node of the last level might not have left and right nodes. Is that correct?  And would the complexity be any different if it is DFS?

super30admin:
@sowmyadvn Nope the complexity (worst case) scenario of DFS would have been O(n) as well for binary tree.

sowmyadvn:
@super30admin  Thanks a lot for clarifying! :) I am having fun coding already!

PR #92, opened 2020-01-30
Completed PreCourse_1. Please Review

vinaypotla17:


super30admin:
Great Job Vinay. Liked the way you commented your code. Add time and space complexities. Why didn't we use a size variable for doing Level Order Traversal. How do we identify that we need not use a size variable?

vinaypotla17:
From what I have understood of the question, as we are using a queue it traverses the whole binary tree and as we are using a linked list we can dynamically add and remove the nodes.

PR #100, opened 2020-03-03

UzmaArzoo:


Exercise_1.java
```diff
@@ -1,38 +1,80 @@
...
+    	int popX =0;
         //Write your code here
+    	if(!isEmpty())
+    	{
+    		 popX = a[top--];
+    		 return popX;
+    	}
+    	else
+    	{
+    		isEmpty();
```
devashishnyati: There's no need to call isEmpty() in the else condition again

devashishnyati:
Good job Arzoo!
In problems 1 and 2, you are calling isEmpty() again in the else conditions. You can avoid that. It will print the output twice.

Exercise_1.java
UzmaArzoo: Alright, got it. I will update it in my code.

Thank you!

PR #102, opened 2020-03-04
Completed precourse1

AroopN:


devashishnyati:
Good job Aroop! Try doing inorder traversal iteratively also!

AroopN:
> What will be time complexity of popping first element from a list. What is under the hood implementation of list in python?

Thank you. Popping the first element from list in python O(n). Using deque from collections library will give us a time complexity of O(1).  Popping the last element is still O(1) in python lists.

PR #105, opened 2020-03-06
tried couple of questions please review

rshah02:


Exercise_1.java
```diff
@@ -5,28 +5,47 @@
...
+         return false;
+     }else{
+         a[++top]=x;
+         return true;
+     }  
     } 
   
     int pop() 
     { 
+        if(top<0){
```
devashishnyati: You can use isEmpty() function here

Exercise_2.java
```diff
@@ -1,40 +1,63 @@
...
-        //Write your code here for the condition if stack is empty. 
+        return root==null;//Write your code here for the condition if stack is empty. 
+    
     } 
   
     public void push(int data) 
     { 
+    	StackNode temp=new StackNode();
+    	if(temp==null) {
+    		System.out.print("\n stack overflow");
```
devashishnyati: If temp is null, stack won't overflow. Stack can overflow if you are implementing it using arrays. 
@super30admin am I right?

devashishnyati:
Good job Rohan!
Please complete the remaining problems also!

Exercise_2.java
super30admin: yup you are right

PR #106, opened 2020-03-07
Completed PreCourse_1 - Please Review

kannan-mrugapphan:


Exercise_1.java
```diff
@@ -5,30 +5,51 @@
...
+    	}
+    	a[++top] = x;
+    	return true;
     } 
   
     int pop() 
     { 
-        //If empty return 0 and print " Stack Underflow"
-        //Write your code here
+    	if(top == -1 || a.length == 0)
```
devashishnyati: You can use isEmpty() function here. Also, you can ignore OR condition.

devashishnyati:
Good job Mrugapphan!!

PR #110, opened 2020-03-10
PreCourse1 - please review 

megrao:
Kindly review  - Python code for all 4 problems

### DIFF
--- a/Exercise_1.py
+++ b/Exercise_1.py
@@ -1,18 +1,29 @@
+Time complexity: O(1) for isEmpty, push, pop, peek. O(n) for size and show.
+Space complexity: O(1) for all
+Executed in Leetcode: No link. Code executed in jupyter notebook.
+Challenges: Not sure about time and space complexity.  Can we use size() function in peek?
+     
 class myStack:
      def __init__(self):
+          self.items = []
          
      def isEmpty(self):
+          return self.items == []
          
      def push(self, item):
+          return self.items.append(item)
          
      def pop(self):
-        
+          return self.items.pop()
         
      def peek(self):
+          return self.items[size(self) - 1]   # Can we do this?
         
      def size(self):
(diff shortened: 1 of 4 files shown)


Exercise_1.py
```diff
@@ -1,18 +1,29 @@
+Time complexity: O(1) for isEmpty, push, pop, peek. O(n) for size and show.
+Space complexity: O(1) for all
```
devashishnyati: Just to clarify, space complexity for each operation here is O(1) but the overall space is O(n) which is used by the stack.

Exercise_1.py
```diff
@@ -1,18 +1,29 @@
...
+Executed in Leetcode: No link. Code executed in jupyter notebook.
+Challenges: Not sure about time and space complexity.  Can we use size() function in peek?
```
devashishnyati: To get the length of a list, in python there's len() function.

Exercise_1.py
```diff
@@ -1,18 +1,29 @@
...
          
      def push(self, item):
+          return self.items.append(item)
          
      def pop(self):
-        
+          return self.items.pop()
         
      def peek(self):
+          return self.items[size(self) - 1]   # Can we do this?
```
devashishnyati: No its won't work.
One of the ways to handle this is return self.items[len(self.items) - 1]

devashishnyati:
Good job Megha!
Very nice explanations! I hope you have a better understanding of these data structures now.
Please add a comment or ask me if I missed any of your questions.

Exercise_1.py
megrao: I have used the len() function inside the size function.

Exercise_1.py
megrao: i have defined a size function with len() function inside it.

PR #111, opened 2020-03-10
precourse 1 MR from Ayesha. Please Review.

asiddiqua004:


Exercise_1.cpp
```diff
@@ -1,52 +1,119 @@
...
-    class Stack s; 
-    s.push(10); 
-    s.push(20); 
-    s.push(30); 
-    cout << s.pop() << " Popped from stack\n"; 
-  
-    return 0; 
-} 
\ No newline at end of file
+/ Time Complexity :
```
devashishnyati: Can you explain the time complexity here?

devashishnyati:
Good job Ayesha!

Exercise_1.cpp
asiddiqua004: Worst case is Time complexity is 0(n) as new space has to be created and data is to be copied in.

PR #115, opened 2020-04-14
Completed sprint precourse1 -- Please review

VenkateswaraRajuDatla:


Exercise_2.java
```diff
@@ -9,36 +9,65 @@
         StackNode(int data)
```
super30admin, line 9: Better to name it ListNode

super30admin:
Good Job Raju.

At top of each file, please write time and space of each abstract function
In which cases we will need to take size variable in Breadth first search and can you write the DFS/recursive code for level order traversal?

PR #119, opened 2020-04-16
Completed PreCourse1- Please review

varshakatariya:


super30admin:
Loved the PR Varsha 

1. Why we didn't use size variable in BFS here. Where all we need to use a size variable

super30admin:
Loved the PR Varsha

Why we didn't use size variable in BFS here. Where all we need to use a size variable

PR #121, opened 2020-04-16

Siyami:


Excercise_2.js
```diff
@@ -1 +1,67 @@
...
+    constructor(val){
+        this.val = val;
+        this.next = null;
+    }
+}
+
+class SinglyLinkedList{
+    constructor(){
+        this.head = null;
+        this.tail = null;
```
super30admin, line 13: Can we do this with just head pointer. Here is an algorithm for you
Push the node the head (that will be last in)
While popping.  Pop it from head itself. 

This way both Push and Pop functions will be O(1) time. However in your case the pop operation is O(n).

To get a better idea, see the PR by Raju (VenkateswaraRajuDatla)

Both insert and

Excercise_2.js
```diff
@@ -1 +1,67 @@
...
+        return this;
+    }
+    pop(){
+        //If Stack Empty Return 0 and print "Stack Underflow"
+        //Write code to pop the topmost element of stack.
+        //Also return the popped element 
+        let cur = this.head;
+        for(let i = 1; i < this.length - 1; i++){
+            cur = cur.next;
+        }
```
super30admin, line 36: This is not time complexity efficient

Excercise_3.js
```diff
@@ -1,2 +1,134 @@
...
+    constructor(){
+        this.head = null;
+        this.tail = null;
+        this.length = 0;
+    }
+    push(val){
+        let newNode = new Node(val);
+        if(!this.head) {
+            this.head = newNode;
+            this.tail = newNode;
```
super30admin, line 20: Try forming this class with just a head pointer and performing operations based on head only.

This is verbose code

Excercise_3.js
```diff
@@ -1,2 +1,134 @@
...
+        this.val = val;
+        this.next = null;
+    }
+}
+
+class SinglyLinkedList{
+    constructor(){
+        this.head = null;
+        this.tail = null;
+        this.length = 0;
```
super30admin, line 14: We need not have length and tail in constructor. Although they don't hurt yet all abstract functions can be implemented without these variables

Excercise_4.js
```diff
@@ -1 +1,52 @@
...
+
+class Node {
+    constructor(value) {
+      this.value = value;
+      this.left = null;
+      this.right = null;
+    }
+  }
+  
+  class BinarySearchTree {
```
super30admin, line 11: We have a binary tree

super30admin:
Please review inline comments

Excercise_2.js
Siyami: Thanks for the suggestions. If you look at Raju's code and if you push let's say 1-2-3 to his LinkedList you will see that linklist becomes this "3->2->1->null" but it should be the opposite "1->2->3->null" so I think it's a hack. Can you please take a look?

super30admin: What makes you think that its hack? 

If I am correct then you are thinking that LinkedList follows of an order of insertion. A new node has to be added to the tail always?

Siyami: It doesn't always have to be added to the end, a new node can be added anywhere but I think push means adding to the tail.

PR #123, opened 2020-04-17
PreCourse_1 -Review comment changes are done-Please Review

raninagare:


Exercise_4.java
```diff
@@ -1,59 +1,108 @@
...
+     * @param key: this is the data we want to insert in a tree
+     */
+    static void insert(Node root, int key)
+    {
+
+        // Do level order traversal until we find
+        // an empty place and add the node.
+        Queue<Node> que=new LinkedList<>();
+        que.add(root);
+        Node temp=new Node(key);
```
super30admin, line 112: Whats the use of temp node here. Can we do it without the temp node?

super30admin (commented):
Good Job
1. Please write time and space complexities in each file and function.
2. Temp Node in BFS is redundant. Also why didn't we use size variable here?

super30admin:
Good Job

Please write time and space complexities in each file and function.
Temp Node in BFS is redundant. Also why didn't we use size variable here?

PR #125, opened 2020-04-20
Completed Pre course1

ashutosh-senapati:


Exercise_1.java
```diff
@@ -1,44 +1,71 @@
+import java.util.*;
+//push - O(1)
+//pop - O(1)
+//isEmpty() - O(1)
+// peek - O(1)
+
 class Stack { 
+    ArrayList<Integer> stack;
```
super30admin, line 8: Avoid using a higher order Data Structure like ArrayList. Use a normal array

super30admin:
Good Job Ashutosh

PR #127, opened 2020-04-23
Completed Design for 1,2,3 (find it difficult to understand Binary Tree)

JacobStephen9999:


iivaishnavii:
Great job, Jacob! Try completing the 4th problem, when binary trees are being discussed.

Also, do mention the time and space complexities for each problem as well

PR #137, opened 2020-05-18
Completed with external assistance

AbhijeetDhillon:
I had partial knowledge to complete the recourse 1 and I looked up on google to complete the rest

### DIFF
--- a/Exercise_1.java
+++ b/Exercise_1.java
@@ -6,28 +6,57 @@ class Stack {
     boolean isEmpty() 
     { 
         //Write your code here 
+        if(a.length<=0)  
+        return true;
+        else
+        return false;
     } 
 
     Stack() 
     { 
         //Initialize your constructor 
+        this.top=-1;
+        ;
     } 
   
     boolean push(int x) 
     { 
         //Check for stack Overflow
         //Write your code here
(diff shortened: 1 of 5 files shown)


Exercise_1.java
```diff
@@ -6,28 +6,57 @@
     boolean isEmpty() 
     { 
         //Write your code here 
+        if(a.length<=0)
```
Lohith-Aswathappa, line 9: when the stack is empty, it returns false which is incorrect. Check again and update the code

Exercise_2.java
```diff
@@ -9,30 +9,56 @@
...
+        }
+        else
+        {
+            int popedItem = root.data;
+            root = root.next;
+            return popedItem;
+        }
     } 
   
     public int peek()
```
Lohith-Aswathappa, line 58: when the stack is empty and I call the peek function, I should be notified that stack is empty by either printing a default value or error message 
but in this case, I get a null pointer exception, rectify it

Lohith-Aswathappa:
Abhijeet, Good work on trying to solve the precourse exercise
Please review the inline comments and add time and space complexities and with comments.

PR #139, opened 2020-05-19
Completed sprint PreCourse_1 in Python -please review

osamamohamedsoliman:


Exercise_1.py
```diff
@@ -1,19 +1,22 @@
...
      def isEmpty(self):
-         
+          if len(self.arr) == 0:
+               return True
+          else:
+               return False
      def push(self, item):
-         
+          self.arr.append(item)
      def pop(self):
```
Lohith-Aswathappa, line 11: Take care of the edge cases while popping and peeking the elements when the stack is empty. 

In this case, I get an index out of bound error

Lohith-Aswathappa:
Osama, Good job on completing the precourse. 
Please review the inline comments and change the code accordingly, also add time and space complexities

Exercise_1.py
osamamohamedsoliman: resolved and pushed

PR #141, opened 2020-05-19
Completed PreCourse1 material in Python - please review

pcvyas:
Didn't add comments for Exercise 3 as the time complexities were already included

### DIFF
--- a/Exercise_4.py
+++ b/Exercise_4.py
@@ -1,38 +1,64 @@
 # Python program to insert element in binary tree  
-class newNode():  
-  
-    def __init__(self, data):  
-        self.key = data 
+class newNode():
+    def __init__(self, data):
+        self.key = data
         self.left = None
         self.right = None
-          
+
+
 """ Inorder traversal of a binary tree"""
-def inorder(temp): 
-  
-    
-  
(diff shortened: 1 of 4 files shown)


Exercise_4.py
```diff
@@ -1,38 +1,64 @@
...
+# Time Complexity : O(n)
+# Space Complexity : O(n) for the queue we used
+# Did this code successfully run on Leetcode : Didn't try it on leetcode, ran it in the PyCharm IDE
+# Any problem you faced while coding this:
+def insert(temp,key):
+    if temp is None:
+        return
+
+    #we find first available spot for the key to insert using level order traversal
+    queue = []
```
Lohith-Aswathappa, line 32: There's a simple mistake in exercise 4, instead of adding the node once, you are adding it all the time, try to check for the error.

Lohith-Aswathappa:
Good job on completing the precourse Priyansh!
Review the inline comments, please.

Exercise_4.py
pcvyas: Thanks for reviewing Lohith! I’ll fix the mistake

PR #142, opened 2020-05-19

akshatasawhney:


Exercise_1.py
```diff
@@ -1,19 +1,50 @@
+"""
+// Time Complexity : for show() - O(n)
```
chiragshahi, line 2: It's a good practice to define what 'n' is according to the problem. eg- Space: O(n) where n is the size of the stack

Exercise_2.py
```diff
@@ -1,14 +1,30 @@
...
     def __init__(self):
+        self.my_stack = None #initializing my_stack as None
         
     def push(self, data):
+        node = Node(data) #create new node 
+        node.next = self.my_stack # attaching the node at the beginning by adding the pre-existing LL in the next variable of the new node
+        self.my_stack = node #setting head of my_stack to point at the new node
         
     def pop(self):
+        #if self.my_stack==None:
```
chiragshahi, line 23: Underflow condition should be checked

Exercise_4.py
```diff
@@ -7,19 +17,53 @@ def __init__(self, data):
         self.right = None
           
 """ Inorder traversal of a binary tree"""
-def inorder(temp): 
-  
-    
-  
-  
+def inorder(temp): #(left,root,right)
```
chiragshahi, line 20: Great work implementing iterative traversal. Try recursive as well as it is easier and shorter.

chiragshahi:
Great job! Good to see find, delete in Q3

PR #143, opened 2020-05-19

meghnasinghal:


Exercise_1.java
```diff
@@ -6,28 +11,58 @@
...
+        
+        
     } 
   
     boolean push(int x) 
     { 
         //Check for stack Overflow
         //Write your code here
+
+        if(top==MAX)  // return false i.e unable to push if the pointer has reached MAX value i.e array has no more space
```
chiragshahi, line 32: if (top+1 == MAX) or you could write if (size() == MAX) after implementing the function size()

chiragshahi:
Good job! Review the inline comment once

Exercise_1.java
meghnasinghal: I was thinking it to be if(top==MAX) since I am starting top at -1 and the array becomes full at MAX-1, so as soon as it reaches MAX, push shouldn't be allowed.

PR #145, opened 2020-05-19

ashiranka:


Exercise_1.java
```diff
@@ -11,19 +11,26 @@
   
     boolean isEmpty() 
     { 
-        if (top < 0) {
+        // To check if stack is empty, check if top element exists  
+        // if not, then stack is empty
+        if (top < 0) {
```
chiragshahi, line 16: For reference purposes:
This if else block can be written in one line as well    return top < 0 ? true : false; 
or even better would be to write                                  return top == -1;     since top cannot go lower than -1

Exercise_2.java
```diff
@@ -1,3 +1,9 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
```
chiragshahi, line 2: It is a good practice to define what 'n' is in the space/time complexity. 
Space Complexity : O(n) where n is the size of the stack

chiragshahi:
Well done. Please check inline comments

PR #146, opened 2020-05-20

GreeeshmaReddy:


Exercise_1.java
```diff
@@ -5,29 +5,53 @@
   
     boolean isEmpty() 
     { 
-        //Write your code here 
+        if(top == -1){
```
chiragshahi, line 8: For reference purposes:
This if else block can be written in one line as well return top < 0 ? true : false;
or even better would be to write return top == -1; since top cannot go lower than -1

Exercise_1.java
```diff
@@ -5,29 +5,53 @@
...
+            top = top-1;    //decrementing the index to pose that stack doesn't contain this element anymore.
+        }
+        return x;
     } 
   
     int peek() 
     { 
-        //Write your code here
+        int x = 0;
+        if(!(isEmpty())){
```
chiragshahi, line 50: if (isEmpty()) {}

Exercise_4.java
```diff
@@ -31,10 +32,34 @@ static void inorder(Node temp)
...
     /*function to insert element in binary tree */
     static void insert(Node temp, int key) 
-    { 
-
-        // Do level order traversal until we find 
-        // an empty place and add the node.  
+    {   
+        Node n = new Node(key);
+        Queue q = new LinkedList(); //Using the queue for keepig the track of left and right nodes, so that we can traverse in a level wise
+        Node dummy = null;   //Using this dummy to mark that current level has reached end.
```
chiragshahi, line 38: Can be done without using this dummy node
Look for level order traversal of binary tree online

chiragshahi:
Good work! Q2 becomes a lot easier if you add nodes to the head of the LL instead of the tail.

PR #147, opened 2020-05-20
Completed Sprint PreCourse1 - Please Review.

shantanu609:


Exercise_1.py
```diff
@@ -1,19 +1,27 @@
+# O(N) time | O(1) Space
```
chiragshahi, line 1: Also define what 'N' is 
Space Complexity : O(n) where n is the size of the stack

chiragshahi:
Good work. Nice to see find, remove in Q3. 
Please check for edge cases(underflow, overflow) in Q1 
Also see inline comments

PR #148, opened 2020-05-20
PreCourse Exercise 

cr21:


Lohith-Aswathappa:
Chirag, Good job on completing the precourse. 
Please add time and space complexities from now on.

PR #150, opened 2020-05-20
Completed PreCourse, please review-Shashank

Shashank001122:


chiragshahi:
Good work. Write time and space complexities as well. Q2 becomes a lot easier if you add nodes to the head of the LL instead of the tail.

PR #155, opened 2020-05-22
Review Submission - Precourse_1

Abhishek1004:


Exercise_4.py
```diff
@@ -1,57 +1,57 @@
...
+			temp.right = newNode(key)
+			return
+		elif temp.right:
+			insert(temp.right, key)
+			return
 
 #I failed to understand the condition to insert a node in the binary tree since there is no ordering as such. Therefore it is inserted as the left of 7
   
-    
+
```
chiragshahi, line 37: We have to traverse the tree in a level order way and insert the node whenever we find a null

Exercise_4.py
```diff
@@ -1,57 +1,57 @@
...
+			insert(temp.right, key)
+			return
 
 #I failed to understand the condition to insert a node in the binary tree since there is no ordering as such. Therefore it is inserted as the left of 7
   
-    
+	
   
-       
+
```
chiragshahi, line 39: refer this code for a level order traversal of a tree    
    { 
        // an empty place and add the node.  
        Queue<Node> q = new LinkedList<>();
        q.add(temp);
        while (!q.isEmpty()) {
            int size = q.size();
            for (int i = 0; i < size; i++) {
                Node curr = q.poll();
                if (curr.left == null) {
                    curr.left = new Node(key);
                    break;
                }
                if (curr.right == null) {
                    curr.right = new Node(key);
                    break;
                }
                if (curr.left != null) q.add(curr.left);
                if (curr.right != null) q.add(curr.right);
            }
        }
    }

Exercise_1.py
```diff
@@ -1,22 +1,32 @@
...
-     def show(self):
-         
+
+	def __init__(self):
+		self.stack = []
+
+	def isEmpty(self):
+	 	return self.size() == 0
+	
+	def push(self, item):
```
chiragshahi, line 9: check for overflow condition

Exercise_1.py
```diff
@@ -1,22 +1,32 @@
...
+		self.stack.append(item)
+
+	def pop(self):
```
chiragshahi, line 12: check for underflow condition

chiragshahi:
Good work! Add space, time complexities 
Attempt Q4 again taking help from the inline comments

PR #159, opened 2020-05-24
Precourse 1 completed

aj7274:


Exercise_1.java
```diff
@@ -5,34 +5,53 @@
   
     boolean isEmpty() 
     { 
-        //Write your code here 
+        if(top< 0){
```
chiragshahi, line 8: better way:
return top == -1;

chiragshahi:
Good work! 
Kindly follow start template and write time, space complexities as well

PR #161, opened 2020-05-25
Completed Precourse_1 - Please Review

DineshNadar95:


Exercise_1.java
```diff
@@ -6,28 +6,60 @@
     boolean isEmpty() 
     { 
         //Write your code here 
+        if(top== -1){
```
chiragshahi, line 9: succinct code:
return top == -1;

Exercise_1.java
```diff
@@ -6,28 +6,60 @@
...
     { 
         //Initialize your constructor 
+        this.top = -1;
     } 
   
     boolean push(int x) 
     { 
         //Check for stack Overflow
         //Write your code here
+        if(top>1000){
```
chiragshahi, line 27: 1) wrong condition: if (top + 1 == MAX)
2) don't hard code values, use MAX instead of 1000

Exercise_4.java
```diff
@@ -35,6 +35,26 @@ static void insert(Node temp, int key)
 
         // Do level order traversal until we find 
         // an empty place and add the node.  
+        Queue<Node> insertNode = new LinkedList<>();
```
chiragshahi, line 38: try to use verbs only in function names for better readability of the code

chiragshahi:
Good job Dinesh! 
Follow start template and write time, space complexities
Check inline comments

PR #163, opened 2020-05-25
Precourse 1 - please review

gdhruv03:


Exercise_1.java
```diff
@@ -1,33 +1,60 @@
...
+ * push() : O(1)
+ * pop() : O(1)
+ * peek() : O(1)
+ *  */ 
 class Stack { 
     static final int MAX = 1000; 
     int top; 
     int a[] = new int[MAX]; // Maximum size of Stack 
   
     boolean isEmpty()
```
Lohith-Aswathappa, line 13: if the stack is empty, it still returns false. Check for this case

Lohith-Aswathappa:
Good job Dhruv!
Check the inline comments

PR #166, opened 2020-06-29
Completed precourse-1 in Java - Please check

Shilpi10:
done in java


iivaishnavii:
Good job, Shilpi. Include time and space complexities as well.

PR #170, opened 2020-07-01
Completed PreCourse_1 in Java

akashkshekarappa:
Please Review.

### DIFF
--- a/Exercise_1.java
+++ b/Exercise_1.java
@@ -6,28 +6,50 @@ class Stack {
     boolean isEmpty() 
     { 
         //Write your code here 
+    	if(top == -1) {
+    		System.out.println("Stack empty");
+    		return true;
+    	}
+    	return false;
     } 
 
     Stack() 
     { 
         //Initialize your constructor 
+    	top = -1;
     } 
   
     boolean push(int x) 
     { 
         //Check for stack Overflow
+    	if(top > MAX-1) {
+    		System.out.println("Stack Overflow");
+    		return true;
+    	}
         //Write your code here
+    	else {
+    		a[++top] = x;
+    		return false;
+    	}
     } 
   
     int pop() 
     { 
         //If empty return 0 and print " Stack Underflow"
+    	if(top == -1) {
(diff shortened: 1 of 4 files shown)


Exercise_1.java
```diff
@@ -6,28 +6,50 @@
...
     Stack() 
     { 
         //Initialize your constructor 
+    	top = -1;
     } 
   
     boolean push(int x) 
     { 
         //Check for stack Overflow
+    	if(top > MAX-1) {
```
samarthjj, line 25: The StackOverFlow would occur at "if(top == MAX-1)". Please go through the logic once again.

Exercise_1.java
```diff
@@ -6,28 +6,50 @@
...
+    		System.out.println("Stack Overflow");
+    		return true;
+    	}
         //Write your code here
+    	else {
+    		a[++top] = x;
+    		return false;
```
samarthjj, line 32: return true in case of a successful push operation and false otherwise.

Exercise_2.java
```diff
@@ -9,37 +9,62 @@
...
     { 	
 	//If Stack Empty Return 0 and print "Stack Underflow"
+    	if(isEmpty()) {
+    		System.out.println("Stack Underflow");
+    		return 0;
+    	}
         //Write code to pop the topmost element of stack.
+    	root = root.next;
 	//Also return the popped element 
+		return root.data;
```
samarthjj, line 49: You will have to store the root node before reassigning it, this way you are losing the pointer and returning the incorrect value.

Exercise_1.java
akashkshekarappa: Thanks for the feedback Samarth.
The overflow occurs if top >= MAX-1 ??

samarthjj:
Akash, Good work on completing the PreCourse1. Please review the inline comments.

Exercise_1.java
samarthjj: Yes.

PR #173, opened 2020-07-01
Completed Precourse 1 solutions

ManishaR1515:


Exercise_1.java
```diff
@@ -1,44 +1,51 @@
...
-    } 
-  
-    int pop() 
-    { 
+        if (top == MAX) {
+            System.out.println("Stack Overflow");
+            return false;
+        }
+
+        a[++top] = x;
```
samarthjj, line 21: This will not work as expected. Stores 1st element at index one, will throw **ArrayIndexOutOfBoundsException** for the insertion of the thousandth element.

Exercise_1.java
```diff
@@ -1,44 +1,51 @@
...
-        //Write your code here
-    } 
-  
-    int peek() 
-    { 
-        //Write your code here
-    } 
-} 
-  
+        if (top < 0) {
```
samarthjj, line 27: The if statement will not guard against operating on an empty stack, top is initialized to 0 and will never be less than 0.

samarthjj:
Good work on completing PreCourse1, Please review the inline comments.
For Exercise_2.java (Implementing stack using a LinkedList), inserting to the head and removing from the head of the list will satisfy the property of a stack and result in a better Time Complexity.

samarthjj:
Good work on completing PreCourse1, Please review the inline comments.
For Exercise_2.java (Implementing stack using a LinkedList), inserting to the head and removing from the head of the list will satisfy the property of a stack and result in a better Time Complexity.

Exercise_1.java
ManishaRana1195: got, will fix it.

Exercise_1.java
ManishaRana1195: The top pointer can go below zero as we keep popping elements. Doesn't matter what it was initialized to when creating the stack.

ManishaRana1195:
Great Suggestion on exercise 2. I will open another pull request.

PR #183, opened 2020-07-02
PreCourse 1 completed, please review!

arrpitsharrma:


Exercise_4.py
```diff
@@ -1,38 +1,60 @@
-# Python program to insert element in binary tree  
-class newNode():  
-  
-    def __init__(self, data):  
-        self.key = data 
+# Time Complexity : not sure
+# Space Complexity : not sure
+# Did this code successfully run on Leetcode : yes
+# Any problem you faced while coding this : had a hard time dealing with tree, I get stuck when I come across trees and linked list for basic things sometimes
```
AbhijeetDhillon, line 4: Don't worry, these topics will be covered in depth once your classes begin. You have had a good approach towards the problem

AbhijeetDhillon:
Good Job Arpit, you have successfully completed Precourse 1.

AbhijeetDhillon:
Good Job Arpit, you have successfully completed Precourse 1.

arrpitsharrma:
I think my solution for question 4 is not correct. I took it as a binary tree instead of a tree. Can you please comment on that?

PR #186, opened 2020-07-04

gadlakha:


Exercise_1.py
```diff
@@ -1,22 +1,38 @@
...
+
      def push(self, item):
+     	self.stack.append(item)
          
      def pop(self):
-        
+         return self.stack.pop()
         
      def peek(self):
+         return self.stack[len(self.stack)-1]
```
samarthjj, line 19: peek will crash for an empty stack, please include edge cases.

Exercise_1.py
```diff
@@ -1,22 +1,38 @@
 class myStack:
      def __init__(self):
-         
+         self.stack = []
+
      def isEmpty(self):
-         
+
+     	 if self.stack is None:
```
samarthjj, line 7: This will never return True, because stack is initialized as an array and will never be a None pointer.

Exercise_3.py
```diff
@@ -17,16 +19,77 @@ def append(self, data):
...
         Search for the first element with `data` matching
         `key`. Return the element or `None` if not found.
         Takes O(n) time.
         """
+        temp=self.head
+        if temp==None:
+            return None
+        while temp!=None:
+            if temp.data==key:
+                return temp.data
```
samarthjj, line 42: Garima, What will the function return If the list exists but the key to find is not in the list?

Exercise_4.py
```diff
@@ -8,13 +9,32 @@ def __init__(self, data):
...
+      inorder(temp.right)    
   
   
 """function to insert element in binary tree """
-def insert(temp,key): 
-  
+def insert(temp,key):
+    #print(temp) 
+    list = []  
+    list.append(temp)
```
samarthjj, line 22: Please include edge case for insertion of root node, when the tree doesn't exist.

samarthjj:
Good work on finishing PreCourse1, Garima. Please always include edge cases.
Review the inline comments and let me know if you need help.

Exercise_4.py
gadlakha: Hi Samarth,

Thanks for reviewing the exercises.
I have updated exercise 4 to include the edge case but I am getting one issue, on printing the inorder traversal:
Result are:(Case where only key =12 is getting inserted)
Inorder traversal before insertion: None
Inorder traversal after insertion: 12  None
As per my understanding None should not be present. Please let me know where I have made the mistake

Thanks in advance.

-Garima

PR #187, opened 2020-07-05
First checkin

sujabala:


Exercise_1.java
```diff
@@ -1,32 +1,46 @@
...
         //If empty return 0 and print " Stack Underflow"
         //Write your code here
     } 
   
+    public Boolean isFull() {
+        return top == MAX - 1;
+      }
     int peek() 
     { 
+    	return a[top-- + 1];
```
samarthjj, line 43: This way you'll lose the top pointer, for peek the top pointer decremented.
"return a[top]" should work just fine.

Exercise_1.java
```diff
@@ -1,32 +1,46 @@
...
+
+    	top = -1;
         //Initialize your constructor 
     } 
   
     boolean push(int x) 
     { 
+    	a[++top] = x;
+    	if(isFull())
+    		return false;
```
samarthjj, line 24: The StackOverflow should be checked before adding. This way even on a full stack the program will add the value and then return False.

Exercise_2.java
```diff
@@ -8,18 +9,31 @@
...
-            //Constructor here 
+        	next.data = data;
         } 
     } 
     
 	
     public boolean isEmpty() 
     { 
+    	if(root == null)
+    		return false;
```
samarthjj, line 20: if root is null, that implies the stack is empty. Shouldn't it be returning true here and false otherwise?

Exercise_3.java
```diff
@@ -22,8 +23,22 @@
...
+    	Node currentNode = new Node(data);
+    	
+    	if(list.head == null)
+    	{
+    		list.head = currentNode;
+    		
+    	}
+    	else {
+    		currentNode.next = list.head.next;
+    		list.head = currentNode;
```
samarthjj, line 37: This isn't correct, please go through the logic again.

Exercise_4.java
```diff
@@ -32,6 +34,27 @@ static void inorder(Node temp)
     /*function to insert element in binary tree */
     static void insert(Node temp, int key) 
     { 
+    	Queue<Node> q = new LinkedList<Node>();
```
samarthjj, line 37: This will never enter the while loop, you never added temp to the queue.
The queue will be empty and the program will terminate.

Exercise_4.java
```diff
@@ -32,6 +34,27 @@ static void inorder(Node temp)
...
+    		}
+    		else {
+    			q.add(current.left);
+    		}
+    		if(current.right == null) {
+    			current.right = new Node(key);
+    		}
+    		else {
+    			q.add(current.right);
+
```
samarthjj, line 53: After adding in both else statements, the loop needs to break. Otherwise the node will be added multiple times at where ever it encounters a null.

samarthjj:
Good work on completing the PR Sujatha. Please review the inline comments and rectify your mistakes. You can text me on Slack if you need any help.

samarthjj:
Good work on completing the PR Sujatha. Please review the inline comments and rectify your mistakes. You can text me on Slack if you need any help.

PR #194, opened 2020-07-06
Completed PreCourse1 -> Please review it. 

SantoshReddyKallala:
Trying to catch up sorry for the delay.


Exercise_1.java
```diff
@@ -1,44 +1,74 @@
...
+            return false;
+        }
+
     } 
   
     int pop() 
     { 
         //If empty return 0 and print " Stack Underflow"
         //Write your code here
+        if(top!=-1 && top<MAX-1){
```
samarthjj, line 35: The program fills the stack until the 999th Index, will pop work if top is 999?

Exercise_1.java
```diff
@@ -1,44 +1,74 @@
...
+            return 0;
+        }
+
     } 
   
     int peek() 
     { 
-        //Write your code here
-    } 
+        return a[top];
```
samarthjj, line 49: what If top is -1? Please add the edge case to guard operating on an empty stack.

samarthjj:
Good work Santosh, Please review the inline comments.
Also for exercise 2 (stack using linked list), adding removing from the head will satisfy the property of a stack and result in a much better runtime.

samarthjj:
Good work Santosh, Please review the inline comments.
Also for exercise 2 (stack using linked list), adding removing from the head will satisfy the property of a stack and result in a much better runtime.

PR #198, opened 2020-07-24
Please review pre-course 1

kakashihatakae:


Exercise_1.py
```diff
@@ -1,19 +1,25 @@
...
+          return len(self.stack) > 0
+
      def push(self, item):
-         
+          self.stack.append(item)
+
      def pop(self):
-        
-        
+          return self.stack.pop()
```
samarthjj, line 12: Please add edge case to prevent from operating on a empty stack.

Exercise_1.py
```diff
@@ -1,19 +1,25 @@
 class myStack:
      def __init__(self):
-         
+          self.stack = []
+     
      def isEmpty(self):
-         
+          return len(self.stack) > 0
```
samarthjj, line 6: isEmpty should return True if size is 0 and not the other way around.

samarthjj:
Good work on completing PreCourse1 Shreyas!

samarthjj:
Good work on completing PreCourse1 Shreyas!

PR #214, opened 2020-08-13
Completed Precourse1 - please review

NithyaKuchadi:


Exercise_1.java
```diff
@@ -1,33 +1,51 @@
...
-        //Initialize your constructor 
+       top = -1;
     } 
   
     boolean push(int x) 
-    { 
+    {
         //Check for stack Overflow
         //Write your code here
+    	if(top < MAX)
```
samarthjj, line 21: if(top < MAX -1), The stack is full at index 999 not 1000.

Exercise_1.java
```diff
@@ -1,33 +1,51 @@
...
+    	{
+    		int x= a[top];
+    		a[top--] = 0;
+    		return x;
+    	}
+    	System.out.println("Stack Underflow");
+    	return 0;
     } 
   
     int peek()
```
samarthjj, line 46: Please add edge case to prevent from operating on an empty stack.

samarthjj:
Nithya, Good work on completing Pre_Course-1, Please review the inline comments.

samarthjj:
Nithya, Good work on completing Pre_Course-1, Please review the inline comments.

Exercise_1.java
NithyaKuchadi: If the stack is empty, top will be -1. If top is not -1, I am accessing stack elements. (This is the edge case. please let me know if I have to add anything)

PR #215, opened 2020-08-13
Completed Precourse1- Please review 

vatsaashwin:


Exercise_4.py
```diff
@@ -8,12 +8,31 @@ def __init__(self, data):
           
 """ Inorder traversal of a binary tree"""
 def inorder(temp): 
-  
-    
-  
-  
+    if temp:
+        inorder(temp.left)
```
kannan-mrugapphan, line 12: Please try to solve it using an iterative approach as well

kannan-mrugapphan:
Good Job

PR #244, opened 2020-09-13
Completed Pre-Course 1 - Please review

meetdave06:


Exercise_1.py
```diff
@@ -1,22 +1,32 @@
 class myStack:
      def __init__(self):
+          self.stack = []
```
kannan-mrugapphan, line 4: You should implement a stack using an array and not use inbuilt methods

kannan-mrugapphan (commented):
Good Job. Include the time and space costs for each operation as well.

kannan-mrugapphan:
Good Job. Include the time and space costs for each operation as well.

PR #246, opened 2020-09-14
Completed PC1 Please Review

Lakshya0744:


Exercise_1.py
```diff
@@ -1,22 +1,39 @@
 class myStack:
      def __init__(self):
-         
+         self.arr=[]
+
```
kannan-mrugapphan, line 4: You should not use in built methods. You should build a stack using an array

kannan-mrugapphan (commented):
Good Job. Include the time and space for each operation as well.

kannan-mrugapphan:
Good Job. Include the time and space for each operation as well.

PR #260, opened 2020-10-08

GayatriHungund81194:


Exercise_1.py
```diff
@@ -1,19 +1,51 @@
+# Time Complexity : O(n)
```
samarthjj, line 1: O(1), All functions have constant time complexity.

Exercise_2.py
```diff
@@ -1,22 +1,50 @@
+# Time Complexity : O(n)
```
samarthjj, line 1: The better way to implement this would be to add and remove from the head, bringing down the Time complexity to O(1).

Exercise_4.py
```diff
@@ -1,23 +1,44 @@
...
+#When I found a space where either left or right is None, I inserted the node in the tree respectively.
 def insert(temp,key): 
-  
-    
-  
-       
+    bfsList = list()
+    bfsList.append(temp)
+    while(len(bfsList)!=0):
+        elem = bfsList.pop(0)
```
samarthjj, line 29: Causes shifting of all elements to the left, affects the time complexity, Please use a queue or dequeue instead.

Exercise_4.py
```diff
@@ -1,23 +1,44 @@
 # Python program to insert element in binary tree  
+# Time Complexity : O(height)
```
samarthjj, line 2: O(h) for inorder.
O(n^2) for Insert, the usage of a queue instead would bring the time complexity down to O(n).

samarthjj:
Good work Gayatri, Please review the inline comments!

samarthjj:
Good work Gayatri, Please review the inline comments!

Exercise_4.py
GayatriHungund81194: Noted

Exercise_2.py
GayatriHungund81194: Noted

Exercise_1.py
GayatriHungund81194: Noted

Exercise_4.py
GayatriHungund81194: Noted.

PR #263, opened 2020-10-08
Precourse1_submission_sujana

sujana99:
Review my submission

### DIFF
--- a/Exercise_1.py
+++ b/Exercise_1.py
@@ -1,20 +1,27 @@
+##Exercise_1 : Implement Stack using Array.
 class myStack:
-     def __init__(self):
-         
-     def isEmpty(self):
-         
-     def push(self, item):
-         
-     def pop(self):
-        
-        
-     def peek(self):
-        
-     def size(self):
-         
-     def show(self):
-         
+    def __init__(self):
+        self.L=[]
+        self.size=0
 
+    def isEmpty(self):
+        return self.L == []
+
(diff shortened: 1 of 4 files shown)


Exercise_1.py
```diff
@@ -1,20 +1,27 @@
...
+        self.size=0
 
+    def isEmpty(self):
+        return self.L == []
+
+    def push(self,item):
+        self.L.append(item)
+        self.size=self.size+1
+
+    def pop(self):
```
samarthjj, line 14: Shouldn't the size variable be decreased when an element is popped?

Exercise_4.py
```diff
@@ -1,38 +1,55 @@
...
+def insert(temp,key):
+    if not temp:
+        root = newNode(key)
+        return
+    L = []
+    L.append(temp)
+
+    while (len(L)):
+        temp = L[0]
+        L.pop(0)
```
samarthjj, line 25: Causes all elements within the list to shift left, O(n) operation. Use a Queue or Dequeue instead.

samarthjj:
Good work Sujana, Please review the inline comments!

samarthjj:
Good work Sujana, Please review the inline comments!

Exercise_1.py
sujana99: Yes. I was in hurry to submit. Will recheck from now on.

PR #268, opened 2020-10-09
Completed Sprint PreCourse_1 - Please Review

ds-algo-akshay:


Exercise_1.java
```diff
@@ -1,44 +1,54 @@
...
+    }
+
+    boolean push(int x) {
         //Check for stack Overflow
         //Write your code here
-    } 
-  
-    int pop() 
-    { 
+        top++;
```
samarthjj, line 19: Increment top within the "else" part of the statement.
If the function is called several time after the stack is full, followed by a call to pop(). pop() will throw an **IndexOutOfBounds** Exception error because top is much greater than MAX.

Exercise_2.java
```diff
@@ -1,52 +1,63 @@
...
+        root= root.next;
+        return result;
+    }
+
+    public int peek()
+    {
         //Write code to just return the topmost element without removing it.
-    } 
-  
+        return root.data;
```
samarthjj, line 46: Add edge cases to prevent from accessing data when root is null.

samarthjj:
Good work here Akshay, Please review the inline comments!

samarthjj:
Good work here Akshay, Please review the inline comments!

Exercise_1.java
ds-algo-akshay: Good catch. Thank you!

PR #272, opened 2020-10-10
completed sprint precourse 1 - please review 

nishnareddy1:


Exercise_1.py
```diff
@@ -1,19 +1,20 @@
 class myStack:
      def __init__(self):
-         
+         self.s=[]
```
kannan-mrugapphan, line 3: You are expected to populate the array manually and not use built-in methods

kannan-mrugapphan (commented):
Good Job. Please include the time and space complexities as well.

kannan-mrugapphan:
Good Job. Please include the time and space complexities as well.

PR #274, opened 2020-10-12
final-python

pradumna123:


Exercise_1.py
```diff
@@ -1,19 +1,22 @@
 class myStack:
      def __init__(self):
```
kannan-mrugapphan, line 2: You are expected to build a stack and not use built in methods

kannan-mrugapphan (commented):
Good Job. Please include the complexities as well.

kannan-mrugapphan:
Good Job. Please include the complexities as well.

PR #282, opened 2020-10-17
precourse1 review needed

maithrreye:


Exercise_1.py
```diff
@@ -1,19 +1,23 @@
 class myStack:
      def __init__(self):
```
kannan-mrugapphan, line 2: You are expected to use an array and come up with a stack and not use built ins

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #289, opened 2020-11-11
Precouse_1 done - Please review

snk95:


Exercise_1.py
```diff
@@ -1,22 +1,33 @@
...
          
      def push(self, item):
+          self.stack.append(item) # time - O(1)
          
      def pop(self):
-        
+          return self.stack.pop() # time - O(1)
         
      def peek(self):
+          return self.stack[-1]
```
samarthjj, line 15: Please add error guarding, will crash if stack empty.

Exercise_2.py
```diff
@@ -5,10 +5,30 @@ def __init__(self, data):
  
 class Stack:
```
samarthjj, line 6: Implement this SLL Stack, by adding and removing to the head instead of tail. O(1) time for all operations.

Exercise_4.py
```diff
@@ -8,12 +8,35 @@ def __init__(self, data):
...
   
 """function to insert element in binary tree """
 def insert(temp,key): 
+    if temp is None:
+        temp = newNode(key)
+        return
+    queue = []
+    queue.append(temp)
+    while len(queue) != 0:      # level order traversal to add new node to tree
+        cur = queue.pop(0)
```
samarthjj, line 29: Use a queue, instead of a list. 
pop(0) causes shifting, compromises runtime.

samarthjj:
Good work Sarvesh, Please review the inline comments.

samarthjj:
Good work Sarvesh, Please review the inline comments.

PR #296, opened 2020-11-12
Precourse -1 Java implementation

ckallianpur:


Exercise_1.java
```diff
@@ -5,29 +5,47 @@
   
     boolean isEmpty() 
     { 
-        //Write your code here 
+        return (top ==0);
```
samarthjj, line 8: Wouldn't the stack be empty at -1, and not 0?

Exercise_2.java
```diff
@@ -1,4 +1,4 @@
-public class StackAsLinkedList { 
+class StackAsLinkedList {
```
samarthjj, line 1: Implement by adding and removing from instead of tail, better Time Complexity.

samarthjj:
Good work Chaitra, Please review the inline comments.

samarthjj:
Good work Chaitra, Please review the inline comments.

PR #307, opened 2020-11-12
Completed Sprint PreCourse1-Please Review

kritagya321:


Exercise_2.py
```diff
@@ -5,10 +5,25 @@ def __init__(self, data):
  
 class Stack:
     def __init__(self):
+        self.head = None
         
     def push(self, data):
+        if self.head == None:
```
samarthjj, line 11: Should be if, else condition, first node gets added twice otherwise.

Exercise_4.py
```diff
@@ -7,13 +7,37 @@ def __init__(self, data):
...
+def insert(temp,key):
+    if temp == None:
+        root = newNode(key)
+    else:
+       q = [] 
+       q.append(temp) 
+ 
+    while (len(q)): 
+        temp = q[0] 
+        q.pop(0)
```
samarthjj, line 27: Use a queue/deque instead. pop(0) causes shifting.

samarthjj:
Good work Kritagya, Please review the inline comments.

samarthjj:
Good work Kritagya, Please review the inline comments.

PR #332, opened 2020-12-17

deepashree5594:


Himanshu-Agarwal93 (commented):
Good improvisation in exercise2. Approach for exercise4 is correct. But looks like you have repeated some code.

Himanshu-Agarwal93:
Good improvisation in exercise2. Approach for exercise4 is correct. But looks like you have repeated some code.

deepashree5594:
I could think of only that approach for Exercise 4. Is this not acceptable?

deepashree5594:
Hi Himanshu, 
Could you please let me know how I can improve my code for Exercise 4? 
Thanks.

PR #371, opened 2021-01-14
Completed PreCourse 1 - Please Review

meghasin:


Exercise_1.py
```diff
@@ -1,18 +1,29 @@
 class myStack:
      def __init__(self):
```
kannan-mrugapphan, line 2: You should use an array and implement stack. You are not recommended to use inbuilts like append() and pop()

kannan-mrugapphan (commented):
Good Job. Also please do include the complexities of each operation as well.

kannan-mrugapphan:
Good Job. Also please do include the complexities of each operation as well.

PR #379, opened 2021-01-14
exercises 1,2,3 please review 

jammulanikhil1:


jammulanikhil1:
Added exercise 4 also

kannan-mrugapphan (commented):
Good Job. Also include the time and space complexities for each operation as well

kannan-mrugapphan:
Good Job. Also include the time and space complexities for each operation as well

PR #385, opened 2021-01-17
Pre-course 1 please check

Hardik3172:
I had some difficulties in double pointers, i was not able to solve it that way, so changed some format

### DIFF
--- a/Exercise_1.cpp
+++ b/Exercise_1.cpp
@@ -10,7 +10,7 @@ class Stack {
 public: 
     int a[MAX]; // Maximum size of Stack 
   
-    Stack() { //Constructor here } 
+    Stack() { top=-1 ;} 
     bool push(int x); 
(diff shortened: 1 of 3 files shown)


Gupta-Keshav-001378461 (commented):
Good Work, Hardik!! Using the single pointer instead of a double pointer to implement stack is fine as it won't affect the time complexity. But I would recommend to brush up about the concepts of pointers whether single, double or triple pointers. If you need some articles to go through just ping me on slack I will share the links.

Gupta-Keshav-001378461:
Good Work, Hardik!! Using the single pointer instead of a double pointer to implement stack is fine as it won't affect the time complexity. But I would recommend to brush up about the concepts of pointers whether single, double or triple pointers. If you need some articles to go through just ping me on slack I will share the links.

PR #399, opened 2021-02-10
Completed Exercise_1.java

SourabhSanghi:
Completed PreCourse Exercise 1

### DIFF
--- a/Exercise_1.java
+++ b/Exercise_1.java
@@ -1,44 +1,75 @@
-class Stack { 
-    static final int MAX = 1000; 
-    int top; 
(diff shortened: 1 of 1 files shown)


Exercise_1.java
```diff
@@ -1,44 +1,75 @@
...
+                      }
+        else {
+            return false ;
+             }
+        }
+
+    Stack1()
+    {
+        //Initialize your constructor
+        top = 0;
```
yahska-ednihs, line 20: Here the stack is initialized to top = 0 but the stack is empty when top = -1. So the stack has no elements when it's initialized but isEmpty will return False.

PR #402, opened 2021-02-10
Precourse1 Done in C++ : Please Review

Yashzaparto:


Exercise_1.cpp
```diff
@@ -21,22 +25,33 @@ bool Stack::push(int x)
 { 
     //Your code here
     //Check Stack overflow as well
+    if(top>=MAX)return false;
+    a[++top] = x;
```
yahska-ednihs, line 29: Please check this line. I am rusty with c++ but I believe here when top is 999 it will not return on line 28. On line 29 it will increment to 1000 and try to access a[1000] which will cause error.

Exercise_1.cpp
```diff
@@ -47,6 +62,13 @@ int main()
     s.push(20); 
     s.push(30); 
     cout << s.pop() << " Popped from stack\n"; 
+    cout << s.isEmpty() <<" Is is empty?\n";
```
yahska-ednihs, line 65: Overall Good Job! Please add comments and time and space complexities and review inline comments.

Exercise_2.cpp
```diff
@@ -19,21 +20,45 @@ StackNode* newNode(int data)
...
 { 
     //Your code here 
+    return root?0:1;
+
 } 
   
 void push(StackNode** root, int data) 
 { 
     //Your code here 
+    StackNode* t = new StackNode;
```
yahska-ednihs, line 30: This implementation will work but not common where you add a new node before the head. Please add comments and time and space complexity

Exercise_3.cpp
```diff
@@ -44,25 +60,57 @@ void append(Node** head_ref, int new_data)
...
+
 }  
   
 // This function prints contents of 
 // linked list starting from head  
 void printList(Node *node)  
 {  
     //Your code here
+    Node* iter = node;
+    if(node == NULL)return;
```
yahska-ednihs, line 109: Good Job! Please add time and space complexities for each operation

Exercise_4.cpp
```diff
@@ -31,9 +37,29 @@ void inorder(struct Node* temp)
 /*function to insert element in binary tree */
 void insert(struct Node* temp, int key) 
 { 
-  
+    
     // Do level order traversal until we find 
     // an empty place.  
+    if(temp == NULL){
+    temp = newNode(key);
+    return;
```
yahska-ednihs, line 45: Good Job! Please add comments, time and space complexities for each operation

PR #430, opened 2021-03-02
 done exe1 & 2

stuti19989:


khushal1198 (commented):
Good work with the First two exercises!

khushal1198:
Good work with the First two exercises!

PR #455, opened 2021-03-14
Please review 1 through 5

Kush-16:


Exercise_1.py
```diff
@@ -1,18 +1,25 @@
 class myStack:
      def __init__(self):
+          self.items = [] 
          
      def isEmpty(self):
+          return
```
sibalnirbhay, line 6: The method expects a `bool` type return value.

Exercise_2.py
```diff
@@ -5,10 +5,26 @@ def __init__(self, data):
  
 class Stack:
     def __init__(self):
+        self.head = None
         
     def push(self, data):
+        if self.head == None:
+            self.head = Node(data)
+
```
sibalnirbhay, line 13: I believe either this check should return or should be left out altogether.

Exercise_2.py
```diff
@@ -5,10 +5,26 @@ def __init__(self, data):
...
         
     def pop(self):
+        pop_ele = self.head
+        self.head = self.head.next
+        pop_ele.next = None
+
+        return pop_ele.data
+
+
```
sibalnirbhay, line 27: What is your Stack is empty and someone calls `pop` on it?

sibalnirbhay (commented):
Add TC and SC to your solutions. I have left a few notes of my questions/concerns about your implementation. Think about 'em!

sibalnirbhay:
Add TC and SC to your solutions. I have left a few notes of my questions/concerns about your implementation. Think about 'em!

PR #465, opened 2021-03-17
Please Review PreCourse1 

rashmi1112:


Exercise_1.py
```diff
@@ -1,19 +1,33 @@
...
-     def pop(self):
-        
-        
-     def peek(self):
-        
-     def size(self):
-         
-     def show(self):
-         
+    def __init__(self):
```
yahska-ednihs, line 2: This implementation does not fulfill the requirements for the assignment. Please use the top variable to keep track of top of stack.

Exercise_2.py
```diff
@@ -1,22 +1,58 @@
...
+
 class Stack:
-    def __init__(self):
-        
+    def __init__(self, head=None):
+        self.head = head
+
     def push(self, data):
-        
+        if self.head is None:
```
yahska-ednihs, line 12: Instead of adding elements at the end, try adding them before head.

yahska-ednihs:
Good Job! Please review inline comments.

PR #466, opened 2021-03-17
Stacks and Linked List Solutions

dp2909:
Hello TA, I have been able to finish the first 3 problems in PreCourse 1.
I am trying to wrap my head around the Binary Tree structures for a long time but haven't been able to do so due to my shortcomings at Recursion.
Please leave your comments regarding the first 3 solutions. Meanwhile, I will try to keep working on the Tree question and submit it as soon as possible. Thanks

### DIFF
--- a/Exercise_1_solution.py
+++ b/Exercise_1_solution.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[2]:
+
+
+# Implementation of Stack using Arrays
+
+class myStack:
+    
+    def __init__(self):     
+        self.items = []
+        self.ssize = 0
+        return     
+         
+    def isEmpty(self):
+        if self.items == []:
+            return True
+        else:
+            return False
+         
+    def push(self, item):
+        self.items.append(item)
+        self.ssize += 1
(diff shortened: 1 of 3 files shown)


Exercise_1_solution.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[2]:
+
+
+# Implementation of Stack using Arrays
+
+class myStack:
```
yahska-ednihs, line 9: This implementation does not satisfy the requirements of the assignment. Please implement stack using top variable

Exercise_2_solution.py
```diff
@@ -0,0 +1,77 @@
...
+        self.next = None
+        
+class Stack:
+    def __init__(self):
+        self.head = None
+        self.size = 0
+        return
+        
+    def push(self, data):
+        new_node = Node(data)
```
yahska-ednihs, line 21: Try adding and removing elements in the beginning rather than end.

yahska-ednihs:
Please review inline comments. You can contact the TAs or the professor to get help on the assignments.

PR #552, opened 2021-05-20
Completed PreCourse-1 - Please review

nikhilwani:


kannan-mrugapphan (commented):
Good Job. For stack implementation, don't use inbuilt methods like append() and pop(). Try to implement through custom pointers

kannan-mrugapphan:
Good Job. For stack implementation, don't use inbuilt methods like append() and pop(). Try to implement through custom pointers

nikhilwani:
Hi @kannan-mrugapphan, thank you for reviewing it. Could you please direct me the custom pointer Python solution or code? I'm also wondering if the two-pointer implementation helps in terms of time complexity. pop() and append() in python are O(1). Are there other benefits of using the two-pointer solution that I missed, or is it just for the learning purpose as it would build the two-pointer implementation logic better?

PR #559, opened 2021-05-20
Completed partially as per my current understanding. Please review

takanju:


Exercise_1.py
```diff
@@ -1,18 +1,30 @@
...
+          self.s = []
          
      def isEmpty(self):
+          if len(self.s) == 0:
+               return True 
+          else:
+               return False
          
      def push(self, item):
+          self.s.append(item)
```
kannan-mrugapphan, line 12: Do not use the inbuilt append and pop methods. Please try to implement it with custom pointers.

Exercise_4.py
```diff
@@ -8,12 +8,21 @@ def __init__(self, data):
           
 """ Inorder traversal of a binary tree"""
 def inorder(temp): 
+    if temp is None:
+        return
```
kannan-mrugapphan, line 12: Please try this iteratively as well

Exercise_3.py
```diff
@@ -17,6 +19,15 @@ def append(self, data):
         Insert a new element at the end of the list.
```
kannan-mrugapphan, line 19: Please try to optimize this with a tail pointer

kannan-mrugapphan (commented):
Good Job. Also do add a few comment blocks as well about the approach

kannan-mrugapphan:
Good Job. Also do add a few comment blocks as well about the approach

PR #630, opened 2021-07-13
Completed Excercise-1 from pre-course-1.

ramdinlokesh910:


sridhar-anish (commented):
Good Job with everything, make sure to write TC and SC

sridhar-anish:
Good Job with everything, make sure to write TC and SC

ramdinlokesh910:
@sridhar-anish -I will make sure to add the time complexity and space complexity in my next commits. Thank you for your feedback.

PR #651, opened 2021-07-17

IAnjalyS:


IAnjalyS:
Hello Sai Teja, I could not complete Exercise 4. I need some more understanding on the binary tree logic.

PR #666, opened 2021-08-08
Completed implementing stack with array

meghasn:
I am getting an addition None as output when I call s.show() and also should the array be printed in the reverse order since it is a stack?


meghasn:
complete implementing stack with linked list

meghasn:
implemented singly linked list.so far working well. please do let me know if there is any correction to be made

PR #714, opened 2021-08-17
Completed Pre-course 1, Please review

shilpagowda812:
Ex-2 successfully ran on replit.com but not on LeetCode.It showed error on this below line and also tried to debug but I couldn't fix it.
do = input("What would you like to do?").split()


sridhar-anish:
> Ex-2 successfully ran on replit.com but not on LeetCode.It showed error on this below line and also tried to debug but I couldn't fix it.
> do = input("What would you like to do?").split()



Try running on an IDE

sridhar-anish:
Good job with everything!

PR #758, opened 2021-09-14
Exercise_1.java

deeppatel7981:


Pinak08 (commented):
Good Job!
Add TC and SC

Pinak08:
Good Job!
Add TC and SC

PR #796, opened 2021-11-16
Completed Precourse 1 - Please review

SaiGowtham-11:


main.cpp
```diff
@@ -0,0 +1,93 @@
+#include <iostream>
```
manojvarmarudraraju, line 1: To avoid O(n) iterations over the linked list for append function try to maintain tail pointer as well, so U can append in O(1)

manojvarmarudraraju (commented):
Good job, please try the other problems as well

manojvarmarudraraju:
Good job, please try the other problems as well

PR #805, opened 2021-11-17
Completed pre-course 1 exercises.

Yadnesh-Kulkarni:
Assuming that each function will have a different runtime, Time Complexity of each function has been mentioned separately.


brunda4:
Good Work!

PR #908, opened 2021-12-27
Added precourse 1 solutions

0v14:


sanjayvarma199 (commented):
Good Job !! Also Please try to perform stack operations without using internal operations of list like "pop" may be you can use a counter to keep track of all pop, peek, push operations.

sanjayvarma199:
Good Job !! Also Please try to perform stack operations without using internal operations of list like "pop" may be you can use a counter to keep track of all pop, peek, push operations.

PR #930, opened 2022-01-21
Precourse1 completed - Please review

amishpapneja:


keshav-Gupta-IS (commented):
In your implementation of stack as a linkedlist if you see your peek function you should also check if the list is empty or not. If it is empty then you should return -1.

keshav-Gupta-IS:
In your implementation of stack as a linkedlist if you see your peek function you should also check if the list is empty or not. If it is empty then you should return -1.

PR #931, opened 2022-01-21
PreCourse1 completed. Please Review.

svkhare:


keshav-Gupta-IS (commented):
In your implementation of stack and stack as linkedlist In your peek function you should check whether stack is empty or not before returning a value.

keshav-Gupta-IS:
In your implementation of stack and stack as linkedlist In your peek function you should check whether stack is empty or not before returning a value.

PR #952, opened 2022-01-29

Renali01:


Exercise_1.java
```diff
@@ -1,46 +1,85 @@
+import java.util.EmptyStackException;
+
+import org.w3c.dom.ls.LSException;
+
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : N/A, worked on test cases
+// Any problem you faced while coding this : throw EmptyStackException instead of returning 0 - is this correct?
+
```
shrivatsav, line 9: IMO, for this problem it is acceptable but ideally, it would have been specified what should be returned ( maybe `null` ).

LinkedList.java
```diff
@@ -1,10 +1,18 @@
-import java.io.*; 
+import java.io.*;
+import java.net.http.HttpHeaders; 
+
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : N/A - worked with different test cases, but only after proceeding failed build - what could be the problem here?
```
shrivatsav, line 6: Please elaborate more on the issue with logs / sample test case which raised this specific issue.

LinkedList.java
```diff
@@ -1,10 +1,18 @@
-import java.io.*; 
+import java.io.*;
+import java.net.http.HttpHeaders; 
+
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : N/A - worked with different test cases, but only after proceeding failed build - what could be the problem here?
+// Any problem you faced while coding this : line 55 - question: return type is LinkedList, how would we return by head?
```
shrivatsav, line 7: `head` would be pointing to the first node of the LinkedList and it is being asked to return the first node of the list after appending.

StackAsLinkedList.java
```diff
@@ -0,0 +1,105 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : N/A, work with test cases
+// Any problem you faced while coding this : N/A
+
```
shrivatsav, line 5: Good Work! Try to implement this using O(1) time and space complexity preserving the Stack rules.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

PR #959, opened 2022-02-07
Precourse-1 completed

kbaibhav:


Exercise_2.java
```diff
@@ -1,38 +1,67 @@
-public class StackAsLinkedList { 
+// Time Complexity :
+    //Push: O(1)
+    //Pop: O(n)
+    //peek: O(1)
+// Space Complexity : O(n)
```
shrivatsav, line 5: Good Work! Please try O(1) solution for `pop` operation as well.

shrivatsav (commented):
Good Work!

Exercise_2.java
kbaibhav: I don't know how to do that in singly linked list. It can be achieved in doubly linked list. Please help me with solution

shrivatsav: Please try to push new items and pop items from the `head` of the stack. This will be O(1).

kbaibhav: Got it thanks, updated the code.

PR #1026, opened 2022-03-24
Precourse1 completed please review

Ramya269:


Exercise_1.py
```diff
@@ -1,24 +1,41 @@
+#Time Complexity : O(1)
```
manojvarmarudraraju, line 1: Also try with static array

manojvarmarudraraju:
Good job!!

PR #1070, opened 2022-04-23
commited precourse1 problems in python

Charishma090619999:
Please review


meghasn (commented):
Good Job! Add TC and SC for exercise 1 and 2 also

meghasn:
Good Job! Add TC and SC for exercise 1 and 2 also

PR #1120, opened 2022-05-21
Completed PreCourse 1 - Please review

Ruthwik12:


Exercise_1.py
```diff
@@ -2,19 +2,23 @@ class myStack:
   #Please read sample.java file before starting.
   #Kindly include Time and Space complexity at top of each file
      def __init__(self):
+       self.l = list()
          
      def isEmpty(self):
-         
+         return len(self.l) == 0
      def push(self, item):
+       self.l.append(item)
```
Sumesh-H, line 10: What if stack overflows?

Exercise_1.py
```diff
@@ -2,19 +2,23 @@ class myStack:
...
          
      def pop(self):
-        
+        return self.l.pop()
```
Sumesh-H, line 13: What if stack underflows?

Sumesh-H (commented):
Good Job Ruthwik! Please address the mentioned comments.

Sumesh-H:
Good Job Ruthwik! Please address the mentioned comments.

Exercise_1.py
Ruthwik12: if len(l) > maxsize:
     return -1

will check for this return value in main and will print("stack is at its capacity, cannot insert").

Exercise_1.py
Ruthwik12: if len(l) == 0:
     return -1

will check for this return value in main and will print("stack is empty and cannot pop from stack")

PR #1165, opened 2022-05-29
PreCourse-1 

soumyaChimakurthy:


Sakshi-2103 (commented):
Great work
Kindly add TC, SC and your approach from the next time

Sakshi-2103:
Great work
Kindly add TC, SC and your approach from the next time

PR #1169, opened 2022-05-31
Completed PreCourse - 1 -> Please Review

mohit9494:


niharikagupta2021 (commented):
Good Job! Please check space complexity for all problems

niharikagupta2021:
Good Job! Please check space complexity for all problems

mohit9494:
If we are designing a data structure, are we supposed to consider Space complexity? 
For example, if we are using a dynamic array (resizing) as our base datastructure.

niharikagupta2021:
yes. You should consider any extra/auxiliary space you are using in your solution

PR #1191, opened 2022-06-16
Pre-course 1 Solutions

pranjaliseth:


Exercise_2.py
```diff
@@ -1,3 +1,8 @@
+# Time complexity for :
+   # push operation = O(n)
```
jeelpatel18, line 2: Push() is also having O(1) time complexity as you are adding node at head itself.

jeelpatel18 (commented):
Well Done Pranjali. Please review the comments.

jeelpatel18:
Well Done Pranjali. Please review the comments.

PR #1192, opened 2022-06-16
PreCourse-1 Vivek Sharma

viveksharma180:


StackAsLinkedList.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public boolean isEmpty() {
+        //Write your code here for the condition if stack is empty.
+        if(root == null){ // comparing whether the root is null or not
+            return true;
+        }
+        return false;
+    }
+
+    public void push(int data) {
```
jeelpatel18, line 47: Actually you have to insert a node at beginning of the linked list as you're implementing stack.

StackAsLinkedList.java
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            while (currNode.next != null) { // to traverse to the last node and then insert node there
+                currNode = currNode.next;
+            }
+            newNode.prev = currNode; // connecting the last and new node.
+            currNode.next = newNode;
+        }
+    }
+
+    public int pop() {
```
jeelpatel18, line 62: You will be removing head node while pop().

StackAsLinkedList.java
```diff
@@ -0,0 +1,84 @@
...
+        StackNode currNode = root; // for reference
+        while(currNode.next != null){ //traversing to the last node
+            currNode = currNode.next;
+        }
+        int result = currNode.data; //storing value in int for returning result
+        currNode.prev.next = null; // deleting the last node
+        return result;
+    }
+
+    public int peek() {
```
jeelpatel18, line 73: Returning top means head node's value

jeelpatel18 (commented):
Well Done Vivek. Please review the comments.

jeelpatel18:
Well Done Vivek. Please review the comments.

StackAsLinkedList.java
viveksharma180: I put a check in push() method, if there is no node, then it will create a node(head) and then insert the value in the head node. I believe it will work the same.

StackAsLinkedList.java
viveksharma180: Here, I am returning the latest element that was inserted into the stack.

StackAsLinkedList.java
jeelpatel18: The thing is like you're implementing stack using LinkedList so all operations push(), pop(), peek() should be done in O(1) time right. So, that's why I'm telling that every node will be inserted at beginning of the linkedlist. So, when you pop() first node(which will be head node) will be deleted and same for peek(). For push also, insert node at head and make new inserted node as head.

jeelpatel18: I think now you can imagine the stack in linked list.

viveksharma180: Thanks for the explanation. I implemented it in O(1) now. Please let me know if there is something which can be improved.

PR #1195, opened 2022-06-16
Completed Sprint PreCourse-1 - Please Review

rsp2210:


daredevilteja (commented):
Good Work! Keep it up

Exercise_3.py
```diff
@@ -17,16 +20,51 @@ def append(self, data):
...
+        elif tmp.next:
+            while tmp.next:                       
+                if tmp.next.data == key:
+                    tmp.next = tmp.next.next
+                    #return key
+                    break
+                tmp = tmp.next
+        else:
+            return "Not in the linkedList"  
+
```
ahmedkareem999, line 61: You can simply add this statement print(f"The element removed from the list is:  {key}") and it will print the removed key. Is this what you meant by "not able to return the key?" You have already achieved your objective of removing the first occurrence of the key in the list.

rsp2210: I tried to return the key when I was removing it, but it gives NoneType object. I got your point, that I should print the key instead and break it.

PR #1216, opened 2022-06-21
Completed Sprint Precourse-Please review my Submission

Butulparveen:


Exercise_1.java
```diff
@@ -1,3 +1,5 @@
+//Time Complexity = O(1)
+//Space Complexity = O(1)
```
daredevilteja, line 2: Please check the mentioned space complexity

Exercise_2.java
```diff
@@ -1,3 +1,5 @@
+//Time Complexity = O(1)
+//Space Complexity = O(1)
```
daredevilteja, line 2: Please check the mentioned space complexity

daredevilteja (changes requested):
Please recheck the mentioned time and space complexities of all the problems

daredevilteja:
Good Work! Keep it up

Exercise_1.java
Butulparveen: Is there any mistake in time complexity as per my knowledge I wrote, if there's any mistake suggest me with the correct answer time and space complexity for wrong answers

daredevilteja: Please try to mention the time and space complexity for each function instead of problem as a whole since this is an implementation problem. If you consider for single functions, then the mentioned TC and SC is correct. Here in your implementation, you are fixing the size of the stack to 1000 but in real world we can store more than 1000 items in a stack. So you can't say surely that SC is O(1).

Exercise_2.java
daredevilteja: Please try to mention the time and space complexity for each function instead of problem as a whole since this is an implementation problem. If you consider for single functions, then the mentioned TC and SC is correct.

PR #1226, opened 2022-06-22
Pre course 1 - please review

tusharpawar611:


niharikagupta2021 (commented):
Good Job! Please add sc and tc for all problems

niharikagupta2021:
Good Job! Please add sc and tc for all problems

PR #1239, opened 2022-07-15
Completed 1,2,3 exercises in pre-coursework 1

Charancherry005:


niharikagupta2021:
Good Job! Please add sc and tc as well

PR #1256, opened 2022-07-18
please check pre course 1

jayachanders30acc:
coded in python


niharikagupta2021 (commented):
Good Job! Please include sc and tc for all problems as well

niharikagupta2021:
Good Job! Please include sc and tc for all problems as well

PR #1269, opened 2022-08-11
Completed Precourse1 - Please review

gowthambalboa:


Exercise_1.py
```diff
@@ -1,24 +1,48 @@
+# Time Complexity : O(n) because of push operation and o(1) otherwise
```
ishika2741, line 1: push in stack is O(1)

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

Exercise_1.py
gowthambalboa: Hello Ishika! Yes, but for the worst case scenario for array implementation, isn't it O(n). 
Refer: https://iq.opengenus.org/time-complexity-of-stack/ You can correct me if I am wrong.

PR #1288, opened 2022-08-13
Precourse-1 completed

khushbukareliya07:


Exercise_2.java
```diff
@@ -1,52 +1,116 @@
...
+    }
+
+    public boolean isEmpty()
+    {
+        //Write your code here for the condition if stack is empty.
+        if(length ==0) return true;
+        return false;
+    }
+
+    public void push(int data)
```
jeelpatel18, line 42: See, You have to implement stack using LinkedList,Time for intert/delete should be O(1) according to stack operations. So you should insert/delete element from head of the list. So then it will be Called as stack with time O(1).

jeelpatel18 (commented):
Well Done. Please review the comments.

jeelpatel18:
Well Done. Please review the comments.

PR #1295, opened 2022-08-14
Completed Sprint PreCourse-1 - Please Review

vedantnib:


Exercise_2.py
```diff
@@ -1,15 +1,52 @@
...
        self.data = data
        self.next = None
  
 class Stack:
     def __init__(self):
-        
+        self.head = None
+        self.second_last = None
+        self.last = None
     def push(self, data):
```
jeelpatel18, line 15: You need to just add/remove element at head of linked list. That will be act as a stack .

jeelpatel18: Which will be in O(1)

jeelpatel18 (commented):
Good work..

jeelpatel18:
Good work..

Exercise_2.py
vedantnib: Oh yes. Slipped out of my mind. Will note it

PR #1309, opened 2022-08-17
Pre course 1 solution

meghadusane97:


Stack.java
```diff
@@ -0,0 +1,83 @@
+// Time Complexity : O(n)
```
jeelpatel18, line 1: O(1) time as you're adding/removing element in constant time so .

LinkedList.java
```diff
@@ -0,0 +1,88 @@
+// Time Complexity : O(1)
```
jeelpatel18, line 1: Here for insertion, time is O(N)

jeelpatel18 (commented):
Good job Megha...

jeelpatel18:
Good job Megha...

PR #1317, opened 2022-08-29
PreCourse1 - Harisha Korapati - Completed

HarishaKorapati:


Exercise_3.py
```diff
@@ -17,16 +20,44 @@ def append(self, data):
         Insert a new element at the end of the list.
```
jeelpatel18, line 20: Here Insert() will add element at end of the list not at head of the list

jeelpatel18 (commented):
Well Done.

jeelpatel18:
Well Done.

Exercise_3.py
jeelpatel18: Done

PR #1322, opened 2022-08-31

jaszaveri43:


Exercise_1.py
```diff
@@ -1,24 +1,39 @@
+#Time Complexity: O(n)
```
daredevilteja, line 1: TC is O(1)

daredevilteja (changes requested):
For this type of problems, please mention TC for every function. Also mention whether it executed successfully in leetcode for every problem

daredevilteja:
Good Work! Keep it up

PR #1336, opened 2022-09-01
PreCourse-1 

samritabakshi:
Question 1 and 2 completed
Question 3 half done

### DIFF
--- a/Exercise_1.js
+++ b/Exercise_1.js
@@ -1,0 +1,39 @@
+var Stack = function (maxSize) {
+    this.arr = [];
+    this.length = 0;
+    this.maxSize = maxSize;
+};
+
+Stack.prototype.push = function (x) {
+    if (this.length < this.maxSize) {
+        this.arr[this.length] = x;
+        this.length++
+    }
+};
+
+Stack.prototype.pop = function () {
+    if (this.length == 0) { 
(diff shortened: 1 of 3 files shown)


Exercise_1.js
```diff
@@ -0,0 +1,43 @@
...
+        this.length++
+    }
+};
+
+Stack.prototype.pop = function () {
+    if (this.length == 0) { 
+        console.log("Stack Underflow") 
+        return -1 
+    }
+    else {
```
jeelpatel18, line 23: You are getting element using index/length in O(1) so TC is constant

Exercise_2.js
```diff
@@ -0,0 +1,78 @@
...
+	this.item = null;
+	this.next = null;
+}
+
+LL2Stack.LinkedStack = function(){
+	var head = null;
+	var size = null;
+
+	this.push = function(item){
+		var node = new LL2Stack.StackNode();
```
jeelpatel18, line 18: You are making Stack, so all operations should be done in O(1) time only right.... So add/delete element at head of the linked list

jeelpatel18 (commented):
Well done! Review the comments.

jeelpatel18:
Well done! Review the comments.

PR #1345, opened 2022-09-04
Completed PreCourse-1 . Please review

sh-ruchi:


daredevilteja:
Good Work! Keep it up. Please recheck the TC and SC for 3rd problem and SC for 1st and 2nd problem

PR #1377, opened 2022-09-30
Update Exercise_1.java

eshwar2807:
Eshwar home work


daredevilteja:
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

PR #1386, opened 2022-10-04
completed sprint precourse1 - please review

srin836:


osandeepreddy (commented):
Good job. TC for stack using array is O(1) in all the operations.

osandeepreddy:
Good job. TC for stack using array is O(1) in all the operations.

PR #1396, opened 2022-10-09
Completed PreCourse-1. Please review!

suprajalvs:


Exercise_1.java
```diff
@@ -1,35 +1,56 @@
+//Time and Space Complexity O(top) and worst case O(MAX)
```
osandeepreddy, line 1: TC and SC for stack using array is O(1)

osandeepreddy:
TC and SC for stack using array is O(1)

PR #1430, opened 2022-10-31
Completed Precourse-1- Review Pending

jatinlingala:
Please review


osandeepreddy (commented):
Good Job. Stack using LinkedList can be done more optimal way by pushing and pop at the head.

osandeepreddy:
Good Job. Stack using LinkedList can be done more optimal way by pushing and pop at the head.

PR #1475, opened 2022-12-15
Pre course problems solved, please review

RajeshDM:
I did not find problems 2 and 3 on leetcode and have done local testing. They work with simple examples.


osandeepreddy (commented):
Time complexity to push and pop from head of linkedList is O(1) and for the data structures that are inside the constructor should not be considered so O(1) space.

osandeepreddy:
Recheck time complexities

PR #1483, opened 2022-12-19
Completed course 1

Shivjavali:


daredevilteja:
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem.

PR #1493, opened 2023-01-12
Completed PreCourse-1 - Please review

akhilreddy619:


Exercise_2.java
```diff
@@ -1,52 +1,128 @@
...
   
+    /**
+     * @return
+     * Checks if root is empty, then prints Stack underflow and returns 0. Else, we iterate through
+     * the list to reach the end of the list. Also, we store the previous element of each iteration
+     * so that once we reach the end, we just update previous node's next as null, thus deleting the
+     * top element. Then previous element will be updated to the curr variable to make it the top element.
+     * TC: O(N) where N is the number of elements in the list.
+     */
     public int pop()
```
osandeepreddy, line 71: Stack insertion and deletion should be done on same side.

osandeepreddy:
Stack insertion and deletion should be done on same side.

Exercise_2.java
akhilreddy619: Yes, both are being done from one side in the code. Insertion is being done at the tail. Even the deletion is being done at the tail. Could you please explain the comment in more detail? Thanks!

osandeepreddy: Adding and deleting form the head would give you O(1) for both push and pop operations.

akhilreddy619: Ok! Got it! In my approach its O(N). I will update and push the changes.

PR #1541, opened 2023-03-13
PreCourse-1 problems completed -- Please Review

rasika-1405:


Exercise_2.py
```diff
@@ -1,32 +1,65 @@
+"""
+Rasika Sasturkar
+Time complexity: O(n), total no. of elements in the linkedlist.
```
osandeepreddy, line 3: Pop operation can be O(1) if we add and delete at head of linkedlist

osandeepreddy:
Pop operation can be O(1) if we add and delete at head of linkedlist

PR #1564, opened 2023-05-14
Completed PreCourse-1 Please Review

DheerajGedupudi:
Related to the last question in the Sample.java file.

Where do I add the comments about my approach?


Rohan-Balakrishna:
Good Work Dheeraj ! You need to mention your approach as comments in your code file itself.

PR #1584, opened 2023-05-19
Arnav Bhatia Pre Course 1 

arnavbhatiaa:


Exercise_1.java
```diff
@@ -1,3 +1,10 @@
+// Time Complexity : O(n)
```
osandeepreddy, line 1: TC is O(1)

osandeepreddy:
TC is O(1)

PR #1612, opened 2023-06-01
Update Exercise_1.py

ashishy7:
Written the code in python for the stack DS


pvk3081 (commented):
please complete the other two problems too

pvk3081:
please complete the other two problems too

PR #1619, opened 2023-06-21

lavanyareddy21:


Exercise_2.java
```diff
@@ -8,7 +8,7 @@ static class StackNode {
   
         StackNode(int data) 
         { 
-            //Constructor here 
+             i = 10; //Constructor here
```
lavanyareddy21, line 11: I have added i =10,  how are we going to use this??

lavanyareddy21 (commented):
Please complete the sprint

ishika2741 (commented):
Good work! Kindly add TC and SC as well.

ishika2741:
Good work! Kindly add TC and SC as well.

Exercise_3.py
```diff
@@ -3,6 +3,8 @@ class ListNode:
     A node in a singly-linked list.
```
lavanyareddy21, line 3: Time and space completes for all functions are O(n) - in worse case ,O(1)

PR #1621, opened 2023-06-21
[Super30-PreCourse1] Completed PreCourse-1, Please review

ghost:
Hi Rohan,
           I made 2 commits to complete all 3 exercises in precourse-1 Please verify.
           
 Suggestion : I think we need to update to filenames for java exercises, because it will be difficult to compile and execute java files when class name is different from filenames.

### DIFF
--- a/Exercise_1.java
+++ b/Exercise_1.java
@@ -1,47 +1,71 @@
-class Stack { 
+class Stack {
     //Please read sample.java file before starting.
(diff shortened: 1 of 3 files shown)


pvk3081 (commented):
Good Job Aditya

pvk3081:
Good Job Aditya

PR #1630, opened 2023-06-22

Ayushichourasia02:


Exercise_1.java
```diff
@@ -7,40 +7,65 @@ class Stack {
   
     boolean isEmpty() 
     { 
-        //Write your code here 
-    } 
+        //Write your code here
+        if(top == 0 ){
```
ishika2741, line 11: if top == -1 , then it is in empty state . You have initialised your top pointer with -1 . If top ==0 then it means there is still one element in stack.

ishika2741 (commented):
Well done! Kindly add Time and space complexity for all the problems as well.

ishika2741:
Well done! Kindly add Time and space complexity for all the problems as well.

Exercise_1.java
Ayushichourasia02: fixed it

PR #1661, opened 2023-07-29
code updated

ramyavallandas:
Review Code


Exercise_1.java
```diff
@@ -1,35 +1,65 @@
-class Stack { 
-    //Please read sample.java file before starting.
-  //Kindly include Time and Space complexity at top of each file
+// Time Complexity : O(1)
```
ishika2741, line 1: Kindly mention the complexity for each function separately like for push() -O(1) , pop() - O(1)

Exercise_1.java
```diff
@@ -1,35 +1,65 @@
...
+// Any problem you faced while coding this : No
+class Stack{
     static final int MAX = 1000; 
     int top; 
     int a[] = new int[MAX]; // Maximum size of Stack 
   
     boolean isEmpty() 
     { 
         //Write your code here 
+        if(top < 0)
```
ishika2741, line 13: If top < 0 i.e. if top == -1 then it that case you need to return true (stack is empty)

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

PR #1670, opened 2023-08-09
completed precourse 1- Please review

fastakash:


Nandha2Tech (commented):
Good Job! Kindly make a note of the feedback
In Exercise_1, isEmpty() can be implemented in O(1) by checking the top variable. (if(top<0) stack is empty). In Pop and peek, IsEmpty() function is used . So the time complexities for those functions will also change. Kindly check.

Nandha2Tech (commented):
Kindly check Exercise2. If root is maintained at the top of the stack push, pop and peek can be achieved in O(1). In Line 60, temp.next.next might end up in NullPointerException.

Nandha2Tech:
Good Job! Kindly make a note of the feedback
In Exercise_1, isEmpty() can be implemented in O(1) by checking the top variable. (if(top<0) stack is empty). In Pop and peek, IsEmpty() function is used . So the time complexities for those functions will also change. 

Kindly check Exercise2. If root is maintained at the top of the stack push, pop and peek can be achieved in O(1). In Line 60, temp.next.next might end up in NullPointerException.

PR #1674, opened 2023-08-26

sakethr16:


Nandha2Tech (commented):
Great Work!
In all three exercises SC would be O(n).

Nandha2Tech:
Great Work!
In all three exercises SC would be O(n).

sakethr16:
Hey,
So the SC is O(n) considering the list that is defined in the main() function right? For the stack functionalities it is going to be O(1) or will it be O(n) either way?

Nandha2Tech:
If we consider the stack space it will be O(n). If the stack space is considered as auxiliary space then SC would be considered as O(1) as no extra space is required for any operation.

PR #1677, opened 2023-08-27
Completed PreCourse-1

jailashah:


jailashah:
Please review

Nandha2Tech (commented):
Great Work! For Ex1 TC would be O(1) and SC would be O(n).
Kindly include TC and SC for other two exercises.

Nandha2Tech:
Great Work! For Ex1 TC would be O(1) and SC would be O(n).
Kindly include TC and SC for other two exercises.

PR #1682, opened 2023-08-27
Update Exercise_1.java

Preyash37:
Solution Exercise_1.java


ishika2741 (commented):
Good work! Kindy submit rest of the problems as well.

ishika2741:
Good work! Kindy submit rest of the problems as well.

PR #1730, opened 2023-09-26
Completed PreCourse-1 

RajendraShivaraya:


ishika2741 (commented):
Good work! Kindly include Tc and Sc for each problem.

ishika2741:
Good work! Kindly include Tc and Sc for each problem.

PR #1753, opened 2023-11-07
Create Exercise_1

aditird:


Nandha2Tech (commented):
Great Work! For pop() check for top < 0 and for push check for top == MAX-1. Kindly include TC and SC and submit other problems also.

Nandha2Tech:
Great Work! For pop() check for top < 0 and for push check for top == MAX-1. Kindly include TC and SC and submit other problems also.

PR #1776, opened 2023-12-18
Completed PreCourse-1 in C++

galibhaskar:


Nandha2Tech (commented):
Great Work! Since prevnode reference is given, InsertAfter in Exercise3 takes O(1) TC.

Nandha2Tech:
Great Work! Since prevnode reference is given, InsertAfter in Exercise3 takes O(1) TC.

galibhaskar:
Thanks for your feedback @Nandha2Tech. Yes it is O(1), I overlooked it.

PR #1790, opened 2023-12-22
Precourse-1 PR Completed

Khushbu-Kareliya:


Nandha2Tech (commented):
Great Work! In Exersice1 Since count is initialized with 0, in push operation use a[count] instead of a[count - 1].

Nandha2Tech:
Great Work! In Exersice1 Since count is initialized with 0, in push operation use a[count] instead of a[count - 1].

PR #1905, opened 2024-06-14
Completed Precourse-1

Joshita-22:
Added the 3 new files for the three exercises.


vyankateshmisar:
Good work Joshita. Please include TC and SC in every problem.

vyankateshmisar:
Good work Joshita. Please include TC and SC in every problem.

PR #1966, opened 2024-09-16
Partially Done Precourse1

charanvarma390:


ishika2741 (commented):
Good work! Submit second problem as well

ishika2741:
Good work! Submit second problem as well

PR #1991, opened 2024-10-20
Done PreCourse-1

Abhinav-Sriharsha:


Erlybird:
Please check the peek method of 1st problem.

Erlybird:
Please check the peek method of 1st and 2nd problem. Firstly , you have to check if the stack is empty and then you have return the value of the element, not it's index

Erlybird:
Please check the peek method of 1st and 2nd problem. Firstly , you have to check if the stack is empty and then you have return the value of the element, not it's index

PR #2004, opened 2024-10-22
Precourse-1 completed

Thaslim:


Erlybird:
Handle empty Stack Stack case in peek and pop methods; In the 2nd problem, the pop operation should be in O(1) while your's is in O(N).

Erlybird:
Handle empty Stack Stack case in peek and pop methods; In the 2nd problem, the pop operation should be in O(1) while your's is in O(N).

Erlybird:
Please do these corrections and re-submit them.

PR #2047, opened 2024-12-24
Solved PreCourse-1 Problems

csk731:


Erlybird:
The first problem stack using Array is incorrect, you are initializing the top with 0, it has be -1. It gives wrong isEmpty() function.

Erlybird:
The first problem stack using Array is incorrect, you are initializing the top with 0, it has be -1. It gives wrong isEmpty() function.

csk731:
Hi @Erlybird , I just checked my first solution. 

My thought process behind the initialization of top with 0 is to represent the next available index in the array where the item has to be pushed. That's why I was checking if top==0 in isEmpty() which says if the next element we have to insert  is at  0 means we don't have any elements in the stack. I tried the various test cases and It worked and provided correct results for all inputs but the approach is slightly different.

Thank you for pointing out this as this is a little less readable in this area than the traditional stack implementation. I will try to make it more clear from next time.

PR #2062, opened 2025-01-17
Precourse 1 exercises in Java

AvaniVaidya:


Nandha2Tech (commented):
Great Work! In Stack As Linked List logic, if root node is maintained as top of the stack then time complexity for push, pop and peek operations can be reduced to O(1).

Nandha2Tech:
Great Work! In Stack As Linked List logic, if root node is maintained as top of the stack then time complexity for push, pop and peek operations can be reduced to O(1).

PR #2066, opened 2025-01-18
Complete PreCourse1

ChiragK22:


ishika2741 (commented):
Please submit only the solutions and the extra files that gets generated. It is very confusing to find your code. If possible raise a new PR.

ishika2741:
Please submit only the solutions and the extra files that gets generated. It is very confusing to find your code. If possible raise a new PR.

ChiragK22:
Yes, I’ve created a new PR with only the solutions. I solved the problems
in Python. Please check it now, and let me know if the issue persists.
Thank you!

PR #2079, opened 2025-01-20
PreCourse1 Submission

babsy3:


Nandha2Tech (commented):
Great Work! In Stack Using Linked List, if a root node is maintained as the last inserted node then time complexity for push and pop can be made to O(1).

Nandha2Tech:
Great Work! In Stack Using Linked List, if a root node is maintained as the last inserted node then time complexity for push and pop can be made to O(1).

PR #2245, opened 2025-06-13
Precourse1 - All 3 problems solved

anuragparla:
This PR contains solutions for Exercise1.py, Exercise2.py and Exercise3.py belonging to the PreCourse-1


super30admin:
The student has provided solutions for three different exercises involving stack and linked list implementations. Here's a detailed evaluation:

Strengths:
1. Correctness: 
   - The stack implementations (both array-based and linked list-based) correctly implement all required operations (push, pop, peek, etc.).
   - The linked list implementation correctly handles append, find, and remove operations.
2. Time/Space Complexity: 
   - The student has accurately analyzed and documented the time and space complexity for each method.
3. Code Structure:
   - The code is well-organized with clear method separation.
   - Comments are provided where necessary.

Areas for Improvement:
1. Exercise_1.py:
   - The show() method returns 0 when empty, which is inconsistent with returning the stack contents when not empty. It should return an empty list for consistency.
   - The pop() method returns 0 on underflow, which might be confusing since 0 could be a valid stack value. Consider raising an exception instead.

2. Exercise_2.py:
   - The peek() method returns 0 when empty, which could be ambiguous. Consider returning None or raising an exception.
   - The pop() method's return value isn't properly type-checked before conversion to int (line 49).

3. Exercise_3.py:
   - The remove() method could be simplified by handling the head case separately before the loop.
   - The find() method returns False when not found, which is inconsistent with the docstring stating it should return None.

General Suggestions:
1. Error handling could be more consistent - consider using exceptions rather than returning special values.
2. Type hints could be added to improve code clarity.
3. Some method docstrings could be more detailed about return values and edge cases.
4. The linked list implementation could benefit from a __str__ method for debugging.

PR #2251, opened 2025-06-16
Completed PreCourse-1

shantanujoshi25:


super30admin:
Exercise_1.py:
- Correctness: The stack implementation appears correct with all required methods (isEmpty, push, pop, peek, size, show). However, the pop method could be more efficient by using list.pop() instead of slicing.
- Time Complexity: Correctly identified as O(1) for most operations except show() which is O(n). The pop implementation is actually O(n) due to slicing, not O(1).
- Space Complexity: Correctly identified as O(n).
- Code Quality: Good overall, but could use docstrings for methods. The isEmpty method could be simplified to `return len(self.stack) == 0`.
- Efficiency: The pop method should be modified to use list.pop() for O(1) time complexity.

Exercise_2.py:
- Correctness: The linked list stack implementation is correct with proper push and pop operations.
- Time Complexity: Correctly identified as O(1) for both operations.
- Space Complexity: Correctly identified as O(n).
- Code Quality: Well-structured and readable. Could benefit from docstrings for methods.
- Efficiency: Good implementation with proper pointer manipulation.

Exercise_3.py:
- Correctness: The singly linked list implementation is mostly correct but has a bug in the remove method where it doesn't break after removing the node, potentially causing errors.
- Time Complexity: Correctly identified for append (O(n)), find (O(n)), and remove (O(n)).
- Space Complexity: Correctly identified as O(n).
- Code Quality: Good structure but needs docstrings. The remove method needs fixing as mentioned.
- Efficiency: The remove method could be optimized by breaking after finding and removing the first occurrence.

General improvements needed:
1. Add docstrings to all methods
2. Fix the pop implementation in Exercise_1.py
3. Fix the remove method in Exercise_3.py
4. Consider adding error handling for edge cases (e.g., popping from empty stack)

PR #2281, opened 2025-07-20
Nikhil Lohar_RN03JUL2025 : Complete the problem using Javascript

nikhil2297:
Completed Pre-course


super30admin:
1. Correctness:
   - Exercise_1.js: The stack implementation using an array is mostly correct, but the `pop` method incorrectly uses `slice` which doesn't modify the array in place. It should simply decrement the top pointer.
   - Exercise_2.js: The linked list stack implementation is correct but inefficient for `pop` operations as it requires traversing the entire list to find the previous node.
   - Exercise_3.js: The linked list implementation is correct for insertion and printing, though the `isEmpty` method takes an unnecessary parameter.

2. Time Complexity:
   - Exercise_1.js: Correctly identified as O(1) for peek/pop and O(1) for push (though push is O(1) amortized, not O(N) as stated).
   - Exercise_2.js: Correctly identified O(1) for push/peek but O(N) for pop due to traversal.
   - Exercise_3.js: Correctly identified O(N) for both insert and print operations.

3. Space Complexity:
   - All implementations correctly identify their space complexities.

4. Code Quality:
   - Some inconsistencies in method parameters (e.g., `isEmpty` in Exercise_3.js takes a parameter while others don't).
   - Exercise_1.js could benefit from better error handling (returning undefined/null instead of 0 for underflow).
   - Exercise_2.js could be optimized by maintaining a pointer to the previous node to make pop O(1).

5. Efficiency:
   - Exercise_1.js: The `pop` method can be simplified by just decrementing top without modifying the array.
   - Exercise_2.js: Consider using a doubly linked list or maintaining a pointer to the previous node for O(1) pop operations.
   - Exercise_3.js: The `isEmpty` method could be simplified to not require a parameter since it's an instance method.

PR #2287, opened 2025-07-21
Done PreCourse-1

nitishred13:


super30admin:
1. Correctness:
   - Exercise_1.java: The isEmpty() method has a logical error - it returns false when the stack is empty and true when it's not, which is the opposite of what's expected. The rest of the stack implementation appears correct.
   - Exercise_2.java: The linked list stack implementation is mostly correct, but the pop() method fails when there's only one element in the stack (root.next would be null, causing prevNode to remain uninitialized).
   - Exercise_3.java: The linked list implementation is correct and handles all cases properly.

2. Time Complexity:
   - Exercise_1.java: All operations are O(1) as expected for an array-based stack.
   - Exercise_2.java: Push and pop operations are O(n) due to traversing the entire list, which is inefficient for stack operations (should be O(1)).
   - Exercise_3.java: Insert is O(n) as it needs to traverse to the end, which is expected for this implementation.

3. Space Complexity:
   - All implementations have appropriate space complexity for their respective data structures (O(n) where n is number of elements).

4. Code Quality:
   - Missing comments about time/space complexity at the top of each file as requested.
   - Exercise_2.java could benefit from better variable naming (e.g., 'root' is typically used for trees, 'head' would be more conventional for linked lists).
   - Some redundant comments that just repeat what the code does.

5. Efficiency:
   - Exercise_2.java's stack implementation could be optimized by maintaining a tail pointer or making it a proper stack with LIFO operations at the head (O(1) operations).
   - The isEmpty() logic in Exercise_1.java needs to be fixed.

PR #2290, opened 2025-07-22
Pre course 1

BhavanaJagadish8:


super30admin:
The student's submission appears to be a README file rather than actual code implementations for the three exercises mentioned (Stack using Array, Stack using Linked List, and Singly Linked List). While the README is clear in stating what needs to be implemented, it doesn't demonstrate any actual coding solution to the problems.

For a proper evaluation, we would need to see:
1. Complete implementations of each data structure
2. Proper class structures with methods for standard operations (push/pop for stacks, insert/delete for linked lists)
3. Error handling for edge cases (like popping from empty stack)
4. Comments explaining the implementation choices
5. Test cases demonstrating functionality

The README itself is well-formatted and clear, but it doesn't address the core programming challenges posed by the problem statements. The student should submit the actual implementations of these data structures to receive proper feedback on their coding skills.

PR #2308, opened 2025-08-14
Completed PreCourse 1 Ex.1, Ex.2 & Ex.3

restacksyj:


super30admin:
The student has provided solutions for three different exercises (Exercise_1.cpp, Exercise_2.cpp, and Exercise_3.cpp) that implement stack and linked list operations. Here's my evaluation:

Strengths:
1. Correctness: 
   - Exercise_1.cpp implements a stack using an array correctly with proper push, pop, peek, and isEmpty operations.
   - Exercise_2.cpp implements a stack using linked nodes correctly.
   - Exercise_3.cpp implements linked list operations (push, insertAfter, append, printList) correctly.
2. Time Complexity: The student has correctly identified the time complexities for each operation in their comments.
3. Error Handling: The solutions handle edge cases like stack overflow/underflow and null checks appropriately.
4. Code Structure: The code is well-organized into classes and functions with clear separation of concerns.

Areas for Improvement:
1. Code Quality:
   - The use of `#include <bits/stdc++.h>` is discouraged as it includes all standard library headers, which increases compilation time unnecessarily.
   - Some variable names could be more descriptive (like 'a' for stack array in Exercise_1.cpp).
   - The comments could be more detailed about the purpose of each function rather than just marking "Your code here".
2. Memory Management:
   - In Exercise_2.cpp, the pop operation deletes the node but doesn't check if the root is null before dereferencing it.
3. Best Practices:
   - The stack implementation in Exercise_1.cpp uses a fixed size array (MAX 1000). A dynamic array or vector would be more flexible.
   - The linked list implementation could benefit from a proper class structure rather than standalone functions.
4. Documentation:
   - The problem descriptions are missing, making it harder to evaluate if the solution fully meets requirements.
   - More detailed comments explaining the logic would improve maintainability.

PR #2326, opened 2025-08-19
PreCourse-1 Submission - V1 (Hemish Veeraboina_RN16AUG2025)

v-hemish:
First PR for Precourse 1

### DIFF
--- a/Exercise_1.py
+++ b/Exercise_1.py
@@ -1,20 +1,48 @@
 class myStack:
   #Please read sample.java file before starting.
   #Kindly include Time and Space complexity at top of each file
-     def __init__(self):
-         
-     def isEmpty(self):
-         
-     def push(self, item):
-         
-     def pop(self):
-        
-        
-     def peek(self):
-        
-     def size(self):
-         
-     def show(self):
+
+  """
+  Time Complexity: 
+  Push: O(1)
+  Pop: O(1)
+  Peek: O(1)
+  isEmpty: O(1)
+  size: O(1)
+  show: O(N) as we return the whole list. We can return the whole list, but better practise to return a copy to avoid mutations. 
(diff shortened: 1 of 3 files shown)


super30admin:
The student has demonstrated a good understanding of data structures with their implementations of stack (both array and linked list versions) and a singly linked list. Here are some specific observations:

Strengths:
1. Correctness: The implementations appear functionally correct for basic operations (push/pop for stacks, append/find/remove for linked list).
2. Time/Space Complexity: The student has correctly analyzed and documented time and space complexities for each operation.
3. Code Structure: The code is well-organized into classes with clear methods.
4. Error Handling: Basic error cases are handled (like popping from empty stack).

Areas for Improvement:
1. Exercise_1.py:
   - isEmpty() can be simplified to `return len(self.stack) == 0`
   - peek() should return None or raise an exception rather than just printing when stack is empty
   - show() should return a copy of the stack (as noted in comments) to prevent external modifications

2. Exercise_2.py:
   - The pop() method should raise an exception rather than returning None when stack is empty
   - Could add a peek() method for consistency with array implementation
   - The comment about time complexity being O(1) is correct, but the header says "Linear O(1)" which is contradictory

3. Exercise_3.py:
   - Typo in find() method ('curr' vs 'cur' in variable naming)
   - Remove() method could return the removed node's data for more utility
   - Could add a __str__ method for easier debugging
   - The time complexity comment says "Linear O(1)" which is contradictory - should be either "Linear" or "O(n)"

General Suggestions:
1. Consider using exceptions rather than print statements for error cases
2. Add docstrings explaining return values
3. Consider adding type hints for better code documentation
4. For the linked list implementation, consider adding a size counter to make size operations O(1)

PR #2332, opened 2025-08-24
Completed Precourse-1

Saideepthi123:


super30admin:
1. Correctness: 
   - Exercise_1.py: The stack implementation using an array is mostly correct, handling basic operations like push, pop, peek, and size. However, the `show()` method could be more efficient by only showing elements up to `self.top` without slicing.
   - Exercise_2.py: The linked list implementation of the stack is correct, with proper handling of push, pop, and peek operations. The `size()` and `show()` methods correctly traverse the list.
   - Exercise_3.py: The `append()` method has a bug where it doesn't return after setting `self.head = node` when the list is empty, causing an infinite loop. The `find()` method doesn't return `None` if the key is not found, and the `remove()` method doesn't break after removing the key, which could lead to incorrect behavior.

2. Time Complexity: 
   - The student has correctly identified the time complexities for most operations. However, the `find()` and `remove()` methods in Exercise_3.py should be O(n) as they traverse the list.

3. Space Complexity: 
   - The space complexity is correctly identified as O(n) for all implementations, which is accurate.

4. Code Quality: 
   - The code is generally well-structured and readable, with good comments explaining the logic. However, there are minor issues like the infinite loop in `append()` and missing edge cases in `find()` and `remove()`.

5. Efficiency: 
   - The array-based stack (Exercise_1.py) could be optimized by dynamically resizing the array when it reaches capacity instead of raising an exception.
   - The linked list implementations (Exercise_2.py and Exercise_3.py) are efficient, but Exercise_3.py needs fixes for correctness.

Saideepthi123:
Thankyou for the feedback, will definitely work on it.

PR #2358, opened 2025-11-10
completed precourse 1

PurvaT-11:


super30admin:
1. Correctness:
   - Exercise_1.py: The stack implementation is correct for basic operations (push, pop, peek, size, show). However, the isEmpty() method should ideally return a boolean (True/False) rather than relying on the truthiness of len() == 0.
   - Exercise_2.py: The stack implementation using linked list is mostly correct, but there's no isEmpty() method which would be useful. The pop() method correctly handles empty stack cases.
   - Exercise_3.py: The linked list implementation has several issues:
     - The append() method will fail if the list is empty (self.head is None) because it tries to access curr.next immediately.
     - The find() method returns None too early (inside the while loop) instead of after the loop completes.
     - The remove() method doesn't handle the case where the key is in the head node.

2. Time Complexity:
   - The student has correctly identified time complexities for most operations.
   - For Exercise_3.py, append() is O(n) as stated, but the implementation is incorrect as mentioned above.

3. Space Complexity:
   - The space complexity analysis is correct for all exercises (O(n) for storing elements).

4. Code Quality:
   - Good use of comments explaining the code.
   - Variable naming is clear and appropriate.
   - Missing docstrings for some methods (especially in Exercise_3.py).
   - Exercise_3.py needs better error handling for edge cases.

5. Efficiency:
   - Exercise_1.py could benefit from returning a boolean in isEmpty().
   - Exercise_2.py could include an isEmpty() method.
   - Exercise_3.py needs significant improvements in handling edge cases and correct implementation of methods.

PR #2367, opened 2025-11-20
[IMP] PreCourse-1

amitmittal117:


super30admin:
Exercise_1.py:
- Strengths: The stack implementation using a list is correct and efficient. Time and space complexity analysis is accurate. Methods are well-named and functionality is clear.
- Areas for Improvement: The pop() and peek() methods print an error message but also return None, which could lead to confusion. Better to raise an exception or return None consistently. The show() method could be renamed to something more descriptive like 'get_stack()'.

Exercise_2.py:
- Strengths: The linked list implementation of stack is correct with proper push/pop operations. Good handling of empty stack case in pop(). Clear user interface for stack operations.
- Areas for Improvement: The pop() method could be modified to handle non-integer inputs more gracefully. The class could benefit from additional methods like peek() and isEmpty() for completeness.

Exercise_3.py:
- Strengths: Basic structure of singly-linked list is correct with proper node class definition.
- Areas for Improvement: Several issues need addressing:
  1. append() fails if list is empty (self.head is None)
  2. find() doesn't check the last node's data
  3. remove() has incorrect logic (checks curr.next.next and curr.next.val instead of curr.next.data)
  4. Missing time/space complexity analysis
  5. Missing test cases to verify functionality
  6. The remove() method doesn't handle the case where head node needs to be removed

General Feedback:
- Good practice of including time/space complexity analysis (though missing in Exercise_3)
- Consider adding docstrings for better documentation
- Include more edge case testing (empty stack/list, single element cases)
- Error handling could be more consistent across implementations

PR #2389, opened 2026-01-13
Completed PreCourse 1

varnika-98:
Implemented stack using Python list. Added time and space complexity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/Exercise_1.py
+++ b/Exercise_1.py
@@ -1,21 +1,29 @@
 class myStack:
-  #Please read sample.java file before starting.
-  #Kindly include Time and Space complexity at top of each file
+  #Time Complexity: push: O(1), pop: O(1), peek: O(1), isEmpty: O(1), size: O(1), show: O(n)
+  #Space Complexity: O(n)
+
      def __init__(self):
+         self.items = []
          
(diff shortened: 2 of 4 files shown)


super30admin:
1. Correctness:
   - Exercise_1.py: The stack implementation appears correct with proper push, pop, peek, isEmpty, size, and show operations. The isEmpty check before pop and peek is good practice.
   - Exercise_2.py: The stack implementation using linked list is mostly correct, but the pop method doesn't handle the case where the stack is empty (it calls isEmpty but doesn't return anything in that case). Also, the isEmpty method is not implemented.
   - Exercise_3.py: The singly linked list implementation is correct with proper append, find, and remove operations. Edge cases are handled well.

2. Time Complexity:
   - The student has correctly identified time complexities for all operations in each exercise.
   - For Exercise_2.py, the isEmpty operation would be O(1) if implemented.

3. Space Complexity:
   - Space complexities are correctly identified for all exercises.

4. Code Quality:
   - Code is generally well-structured and readable.
   - In Exercise_2.py, the isEmpty method should be implemented for completeness.
   - In Exercise_3.py, the docstrings could be placed before the method implementations rather than after for better readability.
   - The test cases in Exercise_3.py are a good addition.

5. Efficiency:
   - The implementations are efficient for the given problems.
   - In Exercise_2.py, the pop method could be simplified by directly returning None if the stack is empty rather than having an empty return.

PR #2406, opened 2026-04-06
chore: Completed precourse-1

pkmaster21:


super30admin:

Exercise_1.py:
This implements a stack using a list. The implementation is straightforward and correct for basic stack operations. However, there are a few points to note:
- The `isEmpty` method could be simplified to `return len(self.arr) == 0`.
- The `peek` method should handle the case when the stack is empty to avoid an IndexError. Currently, it will crash if called on an empty stack.
- The `pop` method should also handle the case when the stack is empty. Currently, it will raise an IndexError with "pop from empty list".
- The time complexity comments are accurate: all operations are O(1) except when the list needs to resize (which is amortized O(1)), and space complexity is O(n).
- The code quality is good: it's readable and well-commented.

Exercise_2.py:
This also implements a stack using a list, but with a command-line interface. However, there's a major issue: the class is named `Stack`, but it uses `self.arr` (which is a list) and the methods are implemented using list operations. This is similar to Exercise_1, but the problem might have intended to use a linked list (given the Node class defined but not used). The Node class is defined but never utilized in the Stack class. This seems like a mistake.
- The Stack class should probably be implemented using the Node class to form a linked list, but here it's using a list. This might not meet the problem requirements if the goal was to implement a stack with linked list.
- The time and space complexity are correctly stated.
- The code handles empty stack in `pop` by returning None, which is good.
- The command-line interface is implemented correctly.

Exercise_3.py:
This implements a singly linked list with append, find, and remove operations.
- The `append` method is O(n) as it traverses to the end, which is correct.
- The `find` method is O(n) and returns the node or None, which is correct.
- The `remove` method has an issue: if the list is empty, accessing `self.head.data` will cause an AttributeError (NoneType has no attribute 'data'). Also, if the key is not found, it does nothing (which is acceptable), but it should handle the empty list case.
- The `remove` method should check if the list is empty first. Also, it sets both `current` and `prev` to `self.head` initially, which might lead to issues. Typically, we initialize `prev = None` and `current = self.head`, then traverse.
- The code quality is good, but the removal logic could be improved for clarity and correctness.

General feedback:
- The student has a good understanding of basic data structures but needs to be careful with edge cases (like empty stacks/lists).
- For Exercise_2, there's a disconnect: the Node class is defined but not used. This might be a requirement of the problem to implement a linked list-based stack.
- The time and space complexity analysis is generally correct.

PR #2408, opened 2026-04-07
Pre-course 1

aishwaryasaripalli:
Completed Pre-course 1


super30admin:
Overall, the student has demonstrated a good understanding of basic data structures (stack and linked list) and their operations. However, there are some issues with correctness, edge cases, and code quality that need to be addressed.

For Exercise_1.py (Stack using deque):
- The solution correctly uses a deque to implement a stack. The operations push, pop, peek, size, and show are implemented with appropriate time complexities.
- However, the `pop` method should handle the case when the stack is empty. Currently, it will raise an IndexError if pop is called on an empty stack. It should return None or raise an exception with a message.
- The `show` method returns a list representation of the stack, which is good for debugging.

For Exercise_2.py (Stack using linked list):
- The implementation is mostly correct. The push and pop operations are efficient (O(1)).
- The `pop` method correctly returns None when the stack is empty.
- The code includes an interactive test loop, which is good for testing.
- However, the student mentions that this code was not run on Leetcode, but it should be tested with more edge cases (e.g., popping from an empty stack).

For Exercise_3.py (Singly linked list):
- The `append` method correctly adds to the end of the list. However, it could be optimized by maintaining a tail pointer to avoid O(n) time for each append. Currently, it is O(n) per append.
- The `find` method returns the node if found, which is acceptable. However, the problem says "return the element", which might be interpreted as the data, not the node. This should be clarified.
- The `remove` method handles the case when the node to remove is the head. However, it does not handle the case when the key is not found (it just does nothing). It might be better to return a boolean or raise an exception if the key is not found.
- The code does not handle edge cases well (e.g., empty list, key not found). The student admits being poor at handling edge cases.

General feedback:
- The student has provided time and space complexity analyses, which is good. However, for Exercise_3.py, the append operation is O(n) per operation, which might be inefficient if many appends are done. Maintaining a tail pointer would make append O(1).
- Code readability is decent, but comments are sometimes excessive (e.g., every line commented in Exercise_1.py). Comments should explain why, not what.
- The student should test edge cases (empty stack, empty list, operations on empty structures, etc.).
- For Exercise_1.py, using a deque is efficient, but the problem might expect a pure stack implementation without using built-in data structures. However, since the problem says "Kindly include Time and Space complexity", and the student did, it might be acceptable.

PR #2410, opened 2026-04-08
precourse-1 solution

indrasena483:
Precourse-1 solution


super30admin:
- This should be an overall assessment of whether the solution is acceptable or needs improvement.

IMPORTANT: The student has provided solutions for three exercises. Evaluate each exercise separately and provide an overall verdict for the entire submission.

Let's begin.

PR #2415, opened 2026-05-29
Implement Stack using Array, Implementing Stack using Linked List and Implement Singly Linked List

anirudhv98:
Completed Precourse-1


super30admin:
**Exercise_1.java (Array-based Stack):**
- Correctness: The implementation is mostly correct. The `push` method properly checks for overflow, and `pop`/`peek` handle underflow. However, there's an off-by-one error: `top` is initialized to 0, but in `push`, you're incrementing first then accessing `a[top]`, which means you're writing to index 1 instead of 0. This causes the first element to be skipped.
- Time Complexity: O(1) for all operations - correct.
- Space Complexity: O(MAX) = O(n) where n is the maximum capacity - correct.
- Code Quality: Good structure with clear comments. The overflow/underflow messages are appropriate.

**Exercise_2.java (Linked List-based Stack):**
- Correctness: The approach of using a sentinel value (Integer.MIN_VALUE) to represent an empty stack is unconventional and could cause issues if Integer.MIN_VALUE is a valid data value. A better approach would be to use `null` to represent an empty stack.
- Time Complexity: O(1) for all operations - correct.
- Space Complexity: O(n) where n is the number of elements - correct.
- Code Quality: The sentinel value approach is a code smell. Using `null` would be more idiomatic and avoid potential data collision issues.

**Exercise_3.java (Singly Linked List):**
- Correctness: The implementation is correct. Insertion at the end and traversal work properly.
- Time Complexity: O(n) for insertion (traverses to end) and O(n) for printing - correct.
- Space Complexity: O(n) for the linked list storage - correct.
- Code Quality: Good structure. The code is readable and follows standard linked list patterns.

**General Observations:**
1. The student should add comments about time and space complexity at the top of each file as instructed.
2. The array-based stack has a critical bug that needs fixing.
3. The linked list stack should use `null` instead of a sentinel value for better design.

PR #2418, opened 2026-06-02
preCourse1 assignment

sudheerkreddy:


super30admin:
Let me evaluate each file:

**Exercise_1.java (Array-based Stack):**
- ✅ Correctly implements basic stack operations (push, pop, peek, isEmpty)
- ✅ Proper overflow/underflow handling
- ⚠️ `peek()` returns `a[top]` but should return `a[top-1]` since `top` points to the next empty slot
- ⚠️ `pop()` returns `a[top--]` but should return `a[--top]` to get the correct element
- ⚠️ `print()` method is used but never defined - should be `System.out.println()`
- ⚠️ Time complexity comment says O(1) for all operations which is correct

**Exercise_2.java (Linked List Stack):**
- ✅ Correctly implements push and pop operations
- ⚠️ The StackNode constructor creates another new StackNode internally (infinite recursion bug!)
- ⚠️ `push()` doesn't return a value but the method signature suggests it should
- ✅ Proper underflow handling

**Exercise_3.java (Linked List):**
- ⚠️ The Node constructor creates a new Node inside itself (infinite recursion bug!)
- ⚠️ `insert()` method has issues with null handling and returns incorrectly
- ⚠️ `printList()` is completely empty - no implementation
- ⚠️ `head` is static but should be instance variable for proper OOP design
- ⚠️ The method returns `LinkedList` but should return `Node`

PR #2442, opened 2026-09-18
Completed Precourse-1

pranati05:


super30admin:
1. **Correctness**: The implementations of the stack using a Python list (Exercise_1.py), the linked-list-based stack (Exercise_2.py), and the singly linked list (Exercise_3.py) are all functionally correct. Each method behaves as expected for typical inputs, and the basic edge cases (empty stack/list) are handled.

2. **Time Complexity**: 
   - Exercise_1.py: `push` is amortized O(1), `pop` is O(1), `peek` is O(1), `size` is O(1), `show` is O(1) (returns reference). All operations are optimal.
   - Exercise_2.py: `push` is O(1), `pop` is O(1). Optimal for a linked-list stack.
   - Exercise_3.py: `append` is O(n), `find` is O(n), `remove` is O(n). These are standard complexities for a singly linked list without a tail pointer.

3. **Space Complexity**: 
   - Exercise_1.py: O(n) for the underlying list.
   - Exercise_2.py: O(n) for the linked nodes.
   - Exercise_3.py: O(n) for the linked nodes.
   All are linear in the number of elements, which is expected.

4. **Code Quality**: 
   - The code is generally readable and well-structured. 
   - However, the problem statement asks to "include Time and Space complexity at top of each file," which the student did not do. This is a missed requirement.
   - In Exercise_2.py, the `pop` method returns `None` when empty, but the calling code checks `if popped is None`. This is fine, but it would be clearer to raise an exception (e.g., `IndexError`) for an empty stack, as is conventional in Python. Returning `None` conflates the empty stack case with a stack containing `None`.
   - In Exercise_3.py, the `remove` method has a potential bug: if the key is not found, `current.next` could be `None`, and accessing `current.next.data` would raise an `AttributeError`. The loop should check `if current.next is None: break` or similar.
   - The `print_list` method in Exercise_3.py prints "None" at the end, which is a nice touch for visualization.
   - The test code at the bottom of each file is helpful for verification but should ideally be under `if __name__ == "__main__":` to prevent execution on import.

5. **Efficiency**: 
   - Exercise_3.py could be optimized by maintaining a tail pointer for O(1) appends, but this is not strictly necessary for the problem.
   - The `show` method in Exercise_1.py returns the internal list, which breaks encapsulation. It would be better to return a copy or an iterator.

